PR #107, opened 2015-01-17
Moved models to top level and converted to return Systems.

moorepants:
- The models.py file is not at the root of the pydy package so that all
  subpackages and modules can make easier use of them.
- The mass spring damper model has been converted to a multi mass model,
  similiar to the n-link pendulum.
- Priveate methods were added to System to output the stuff needed for the
  codegen classes. (Ideally in the future, they would just accept Systems, or
  vice versa).


moorepants:
@pydy/pydy-developers How does this look?

moorepants:
@pydy/pydy-developers Thanks for the comments, this is going in.

Issue #109, opened 2015-01-20
Run Travis tests against new and old sympy


moorepants:
We should at least run the tests against the oldest version of sympy we want to support and the HEAD of sympy's master branch.

PR #111, opened 2015-01-20
Support for SymPy 0.7.6, fixed #105 and #109.

### LINKED ISSUE #105: Private methods were removed from the *Methods classes in SymPy 0.7.6
For example:

```
AttributeError                            Traceback (most recent call last)
<ipython-input-2-857534b86bc3> in <module>()
----> 1 from solution.simulation import *

/home/moorepants/Conferences/PYCON2014/pydy-tutorial-pycon-2014/notebooks/solution/simulation.py in <module>()
     66              specifieds={'symbols': specified,
     67                          'values': array([0.0, 0.0, 0.0])},
---> 68              initial_conditions=x0)
     69 
     70 # Generate RHS Function

/home/moorepants/src/pydy/pydy/system.pyc in __init__(self, eom_method, constants, specifieds, ode_solver, initial_conditions, times)
     90 
     91         # TODO : What if user adds symbols after constructing a System?
---> 92         self._constants_symbols = self._Kane_constant_symbols()
     93         self._specifieds_symbols = self._Kane_undefined_dynamicsymbols()
     94 

/home/moorepants/src/pydy/pydy/system.pyc in _Kane_constant_symbols(self)
    515 
    516         """
--> 517         constants = list(self.eom_method._find_othersymbols(
    518             *self._Kane_inlist_insyms()))
    519         constants.remove(sym.symbols('t'))

AttributeError: 'KanesMethod' object has no attribute '_find_othersymbols'
```

moorepants:
- Modified System with conditionals to check for pre and post SymPy 0.7.6 and
  then use the appropriate methods on KanesMethod.
- Added matrix of tests for both SymPy 0.7.4.1 and SymPy HEAD of master to
  Travis.
- Cleaned up some docs in system.py.
- Fixed tests so that order of symbols is ignored.

### DIFF
--- a/pydy/system.py
+++ b/pydy/system.py
@@ -31,8 +31,8 @@ arguments or via the attributes::
     sys.constants = {symbol('m'): 5.0}
     sys.integrate()
 
-To double-check the constants, specifieds, states and times in your problem, look
-at these properties::
+To double-check the constants, specifieds, states and times in your problem,
+look at these properties::
 
     sys.constants_symbols
     sys.specifieds_symbols
@@ -49,10 +49,17 @@ you must call ``generate_ode_function`` on your own::
 
 """
 
-import sympy as sym
+import sympy as sm
+from sympy.physics.mechanics import dynamicsymbols
 from scipy.integrate import odeint
 
 from .codegen.code import generate_ode_function
+from .utils import sympy_equal_to_or_newer_than
+
+SYMPY_VERSION = sm.__version__
+
+if sympy_equal_to_or_newer_than('0.7.6'):
+    from sympy.physics.mechanics.functions import find_dynamicsymbols
 
 
 class System(object):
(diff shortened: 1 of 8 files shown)


moorepants:
@pydy/pydy-developers This is ready to go in as far as I can tell. Reviews welcome.

pydy/system.py
```diff
@@ -545,7 +578,15 @@ def _Kane_constant_symbols(self):
         TODO temporary.
 
         """
-        constants = list(self.eom_method._find_othersymbols(
-            *self._Kane_inlist_insyms()))
-        constants.remove(sym.symbols('t'))
+        from_eoms, from_sym_lists = self._Kane_inlist_insyms()
+        if cmp(parse_version(SYMPY_VERSION), parse_version('0.7.6')) > -1:
```
jcrist: This is repeated frequently throughout this patch. Perhaps this could be turned into a function? Or better yet, evaluated once, and stored in a global somewhere?

moorepants: Yeah, I had something like that in my first shot and then changed it to this because it seemed simpler. I guess the function would be either `greater_than_076()` or `less_than_076`. But if we have new changes in SymPy 0.7,7, then I will need new functions for picking out the version intervals. I guess I thought this was better than starting down that road.

jcrist: It's not so bad, it just could be cleaner. Perhaps you could do something like this, if you wanted:

```
def newer_than(version):
    return cmp(parse_version(SYMPY_VERSION), parse_version(version)) > -1
```

Then this line would be:

```
if newer_than('0.7.6'):
    stuff...
```

moorepants: That's a good suggestion. Will change.

moorepants:
Merging.

Issue #14, opened 2014-04-10
Skip the solve step if the mass matrix is identity or it is not provided


moorepants:
_Issue by **[moorepants](https://github.com/moorepants)** from Wednesday Feb 12, 2014 at 22:57 GMT_
_Originally opened as https://github.com/pydy/pydy-code-gen/issues/21_

----

Right now the numpy.linalg.solve step is a slow down for any of these methods.

If the user solves the for M^-1 * F analytically then we don't need that. This can speed up things right now.

Later on we need to push the solve step to a low level in the generators so there is no python overhead. I'm not sure that the full mass matrix and forcing vector need to be used like this either, especially when the q dots equal the u's, then there is no reason to solve the full system. Finally, we need to do some learning and thinking about the Jain method in which you supposedly don't even have to form the mass matrix and do this solver.

moorepants:
Addressed in PR #113.

jcrist:
closed by #113

Issue #16, opened 2014-04-10
Solve dynamic and kinematic odes separately.


moorepants:
_Issue by **[moorepants](https://github.com/moorepants)** from Sunday Mar 30, 2014 at 03:19 GMT_
_Originally opened as https://github.com/pydy/pydy-code-gen/issues/25_

----

Right now the lazy thing is done: sovle(M, f) where M is the full mass matrix (for dynamical and kinematical equations of motion).

moorepants:
Addressed in PR #113.

jcrist:
Closed by #113

Issue #15, opened 2014-04-10
cython generator can create lots of files


moorepants:
_Issue by **[moorepants](https://github.com/moorepants)** from Wednesday Feb 12, 2014 at 23:29 GMT_
_Originally opened as https://github.com/pydy/pydy-code-gen/issues/22_

----

These should be in a hidden directory and some thought about how to delete ones that are no longer useful or are identical needs to happen.

moorepants:
Addressed in PR #113.

jcrist:
Closed by #113

PR #113, opened 2015-01-29
Adds more flexible code generators for rhs functions that result in faster eval times.

moorepants:
This is pretty close to ready. I've put together a better designed code generation system for the guts of the `generate_ode_function` function. I plan to have this replace everything that exists in `pydy/codegen/code`. I think I can make it completely backwards compatible too, but I may propose a slight backwards incompatibility.

Backwards incompatibilities:

- `pydy.codegen.code.CythonGenerator` is removed. I doubt anyone used this, so I didn't worry about keeping it around. Use `pydy.codegen.cython_code.CythonMatrixGenerator` instead.

Deprecations:
- The `pydy.codegen.code` module will be removed come 0.4.0.
- The generated rhs functions will not accept `args={'constants': ..., 'specified': ...} come 0.4.0.

Fixes issues: #14, #15, #16

- [x] Separated the C and Cython code gen and wrapping into two new modules.
- [x] Generalized the C and Cython code generators to support any number of input arrays and output matrices.
- [x] Developed an abstract base class ODEFunctionGenerator class that makes use of mixins to create the correct rhs functions for three different types of user input: (1) user provides only the rhs, (2) user provides the full mass matrix and accompanying rhs, and (3) user provides mass matrix and rhs for the udots plus the qdots rhs. It is now relatively easy to create subclasses of this to provide more backends.
- [x] Created a CythonODEFunctionGenerator subclass.
- [x] Add support for specified inputs to the ODEFunctionGenerator and subclasses.
- [x] Create a LambdifyODEFunctionGenerator subclass.
- [x] Create a TheanoODEFunctionGenerator subclass.
- [x] Fix bug in SymPy 0.7.4 where Matrix doesn't have free_symbols
- [x] Fix bug in Cython generator that gets confused loading different modules in a row in the compilation step.
- [x] The generated RHS function should support a constant map dict or an array of correct length.
- [x] Add docstrings to the generated rhs functions.
- [x] Allow the args dict to be supplied to the new rhs functions.
- [x] Replace guts of `generate_ode_function` with new code.
- [x] Check why C code doc string isn't always created.
- [x] Switch order of inputs/outputs in CMatrixGenerator and CythonMatrixGenerator to match lambdify.
- [x] Move import errors for optional dependencies into the new respective ODEFunctionGeneratorClasses.
- [x] Make sure warnings fire for old stuff.
- [x] Get it working on SymPy 0.7.4.1
- [x] Support passing in a Generator object to `generate_ode_function`.
- [x] Add option to specify which type of inputs you want in rhs so that you can bypass all the slow arg parsing.

Other enhancements I plan to get to in the future:
- Support creating a function for the Jacobian of the rhs.
- Support functions of time in the ODEs.
- Create Generator class for Octave and Matlab. The Octave one can be wrapped with oct2py.
- Support output equations: both those that share evaluation at each time step and those that can be evaluated after the simulation.
- Add on disk caching for generated functions with https://github.com/bjodah/pycompilation. This can also replace our compilation code.
- Make the CSE optional.
- Support constants=None for EoMs that have numerical values subbed in for all of the constants. You could avoid any of the numerical eval of the constants this way and probably speed things up.

@pydy/pydy-developers I'd love feedback on this, as it is a big change. I'd like to hear what you think of the design and I always fret over function, class, and call signature arg names so thoughts on those are appreciated too.


moorepants:
@chrisdembia @jcrist @gilbertgede @hazelnusse @oliverlee 

This is ready to go in. I'd really love some feedback on it. Thanks.

chrisdembia:
I was a little confused by what `Function` in `ODEFunctionGenerator` meant. It seems that this class provides the RHS of ODE's? `ODERHSGenerator`, or `ODERightHandSide`, or `CanonicalODE`?

moorepants:
@chrisdembia The purpose of the class is to generate a function that evaluates the right hand side of the ODEs. We had a conversation a while back that resulted in us naming the main function in the code module to `generate_ode_function`. So `Function` is in the names of the classes and modules because of that.

I'm having trouble finding the earlier naming conversation. This is commit when it was changed: https://github.com/pydy/pydy/commit/e6025ab2e553467dd519d2aeb8223029af19ad69

moorepants:
Here is the conversation: https://github.com/pydy/pydy-code-gen/issues/8

`ODERHSGenerator` along with `LambdifyODERHSGenerator`, etc seems fine to me. It'll shorten things up with no loss in clarity.

chrisdembia:
Ah okay. Sorry for forgetting the history. I think ODEFunctionGenerator is fine though; ODERHS could be confusing.

pydy/codegen/ode_function_generator.py
```diff
@@ -0,0 +1,705 @@
...
+        f = self._theanoize(outputs)
+
+        def eval_arrays(*args):
+            vals = map(np.asarray, np.hstack(args))
+            return tuple([np.squeeze(o) for o in f(*vals)])
+
+        self.eval_arrays = eval_arrays
+
+
+def generate_ode_function(right_hand_side, coordinates, speeds, constants,
```
jcrist: I assume this is the intended entry point for the user, and all the above classes are just implementation?

pydy/codegen/ode_function_generator.py
```diff
@@ -0,0 +1,705 @@
...
+        The method used for generating the numeric right hand side.
+
+    Returns
+    -------
+    rhs_func : function
+        A function which evaluates the derivaties of the states. See the
+        function's docstring for more details after generation.
+
+    """
+    generators = {'lambdify': LambdifyODEFunctionGenerator,
```
jcrist: Perhaps this could be stored in a global, and a `register` function added? Then a user could create their own generator, and say `register("my_generator", MyGenerator)`, which adds it to the global lookup? This may not be a feature that's ever used/even needed. Just a thought.

pydy/codegen/ode_function_generator.py
```diff
@@ -0,0 +1,705 @@
...
+
+        self._set_eval_array(self._cythonize(outputs, self.inputs))
+
+
+class LambdifyODEFunctionGenerator(ODEFunctionGenerator):
+
+    def _lambdify(self, outputs):
+        # TODO : We could forgo this substitution for speed purposes and
+        # have lots of args for lambdify (like it used to be done) but there
+        # may be some limitations on number of args.
```
jcrist: I'm not sure if python incurs a penalty for number of args. I know numpy `ufuncs` do, but that has to do with their C-api.

pydy/codegen/ode_function_generator.py
```diff
@@ -0,0 +1,705 @@
...
+        subs = {}
+        vec_inputs = []
+        if self.specifieds is None:
+            def_vecs = ['q', 'u', 'p']
+        else:
+            def_vecs = ['q', 'u', 'r', 'p']
+
+        for syms, vec_name in zip(self.inputs, def_vecs):
+            v = sm.DeferredVector(vec_name)
```
jcrist: This is going away. Please don't use this. If `MatrixSymbol` can't do it, this should be fixed.

pydy/codegen/ode_function_generator.py
```diff
@@ -0,0 +1,705 @@
...
+    the correct shape.
+
+    The specified inputs are, in order:
+{specified_list}\
+"""
+
+    @staticmethod
+    def _deduce_system_type(**kwargs):
+
+        if kwargs.pop('coordinate_derivatives') is not None:
```
jcrist: I'm a little confused by this. Are these user set kwargs? They appear to work as long as their defined, so I could pass in `coordinate_derivatives=True` or `coordinate_derivatives=False`, or `coordinate_derivatives="yay duck typing"`, and get the same result each time?

What is the meaning of the different options? Can they contradict each other?

pydy/codegen/ode_function_generator.py
```diff
@@ -0,0 +1,705 @@
...
+
+        def rhs(*args):
+            # args: x, t, p
+            # args: x, t, r, p
+
+            args = self._parse_old_style_extra_args(*args)
+
+            args = self._parse_constants(*args)
+
+            if self.specifieds is not None:
```
jcrist: Since this is always checked, perhaps this check could be moved inside `self._parse_specifieds`?

pydy/codegen/ode_function_generator.py
moorepants: The main entry point is this function, but `Lambdify/Cython/TheanoODEFunctionGenerator` can be too. This function is mainly in place to mirror the function that I removed and follows the SymPy code printers, where a function version is available for the classes.

pydy/codegen/ode_function_generator.py
moorepants: That's easy enough. I can add it.

pydy/codegen/ode_function_generator.py
moorepants: I think someone recently reported this happening to SymPy with a huge number of args in lambdify. There may be a 255 character limit (maybe only in Python 3?).

pydy/codegen/ode_function_generator.py
moorepants: As far as I know MatrixSymbol doesn't work with lambdify yet. Yes it should, but PyDy supports SymPy 0.7.4.1+ right now, so I'm going to keep this in. I wish DeferredVector worked for theano_function too.

pydy/codegen/ode_function_generator.py
jcrist: Actually, after more thought, it may be better to just allow `generator` to be a custom generator object. No globabl registry, seems simpler (perhaps).

pydy/codegen/ode_function_generator.py
jcrist: Mmk. I'll try to get lambdify working with MatrixSymbol this week. Then we can close [this](https://github.com/sympy/sympy/issues/6788), as Deferred Vector is only used in lambdify for this purpose.

pydy/codegen/ode_function_generator.py
moorepants: The generators support three versions of the EoMs:

1. x' = f(x, t, r, p)
2. M(x, p) * x' = f(x, t, r, p)
3. M(q, p) * u' = f(q, u, t, r, p); q' = g(q, u, t)

This code just determines which of the above three you likely passed in. If you pass in garbage (not matrices), then you'll get an error. I could make explicit checks here for what each arg is supposed to be, but right now it is more or less "garbage in, garbage out".

pydy/codegen/ode_function_generator.py
moorepants: Yeah, I don't like how I've defined everything here so much. It also takes way too much time to run these parsing functions to support all the different args in each rhs eval. I'll clean it up.

pydy/codegen/ode_function_generator.py
moorepants: Good idea.

pydy/codegen/ode_function_generator.py
jcrist: Ah, that makes sense. Just curious, looks fine to me as is.

pydy/codegen/ode_function_generator.py
moorepants: Ok, thanks. If you do I'll support it, but I'll have to support both unless we want to up the SymPy dependency version for PyDy.

moorepants:
@pydy/pydy-developers 

Ok, I've addressed the original comments in the latest commits. I also re-factored significantly so that the rhs evaluation can be optimized for speed and the Mixins are gone. I'm ready to merge this. If anyone has any more comments, please make them in the next 24 hrs.

Thanks!

Issue #115, opened 2015-02-06
Travis is failing with SymPy HEAD of master.


moorepants:
https://travis-ci.org/pydy/pydy/jobs/49676673

Not sure why this passed on the PR branch and now it is failing on master.

moorepants:
Ok, I see what is going on. The latest setuptools, must have just been released, is now enforcing PEP 440. SymPy's development version number is `0.7.6-git` which is now not a valid version number according to PEP 440. So when we have the specification `sympy>=0.7.4.1` and have sympy `0.7.6-git` installed it thinks that it needs to upgrad because `0.7.6-git` is treated as an older version that `0.7.4.1`.

One fix is to pin setuptools to the previous version (or maybe it is pip that needs to be).

This is very annoying. They are probably breaking thousands of builds with this change.

moorepants:
Looks like the failing run has this pip version:
`pip 6.0.7 from /home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages (python 2.7)`
But the last passing pip version is:
`pip 1.5.6 from /home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages (python 2.7)`

Looks like Travis just did a major upgrade to pip that is installed with their venv!

moorepants:
But it doesn't look like they've officially upgraded to pip 6+: https://github.com/travis-ci/travis-ci/issues/3060

moorepants:
Ok, this may not be that an older version of SymPy is getting installed but I can't figure it out. The tests pass just fine on my local machine. Can others please run the tests locally with SymPy HEAD of master?

@pydy/pydy-developers 

The error has to to with changes in SymPy that were made in 0.7.6 when Jim refactored KanesMethod. That class not outputs matrices for a bunch of things instead of lists.

moorepants:
I think I get it! Once setuptools and pip are upgrade (as has happened on Travis), then the new function we just created `sympy_equal_to_or_newer_than` is not working correctly because it relies on `from pkg_resources import parse_version` which is from setuptools. `parse_version` now puts 0.7.6-git before 0.7.6!

Need to think of a work around.

paulproteus:
Can you use

`0.7.6.post0.dev`

as the version number, perhaps?

I'm basing this on a reading of https://www.python.org/dev/peps/pep-0440/#developmental-releases which seems to suggest that would work.

moorepants:
[...]

PR #119, opened 2015-02-06
Modified .travis.yml to install pip<6, setuptools<8 to avoid PEP 440 errors.

moorepants:
Fixes issue #115.


moorepants:
This is just some travis fixes, I'm merging so our tests will start to pass again.

PR #140, opened 2015-03-08
Adjust System to support the new generate_ode_function. Fixes #116

### LINKED ISSUE #116: Adjust System to support the new generate_ode_function.
System is setup to use the now deprecated function `generate_ode_function` from the `pydy.codegen.code` module. It needs to be changed to use `pydy.codegen.ode_function_generators.generate_ode_function`.

jinankjain:
  commit message.
  in the unit test.)
  Notes](https://github.com/pydy/pydy/tree/contributing#release-notes).
  follow deprecation cycles.)

### DIFF
--- a/pydy/system.py
+++ b/pydy/system.py
@@ -48,12 +48,13 @@ you must call ``generate_ode_function`` on your own::
     sys.integrate()
 
 """
-
+import warnings
 import sympy as sm
 from sympy.physics.mechanics import dynamicsymbols
 from scipy.integrate import odeint
 
-from .codegen.code import generate_ode_function
+from .codegen.code import generate_ode_function as old
+from .codegen.ode_function_generators import generate_ode_function
 from .utils import sympy_equal_to_or_newer_than
 
 SYMPY_VERSION = sm.__version__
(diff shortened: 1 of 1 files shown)


pydy/system.py
```diff
@@ -462,18 +463,37 @@ def generate_ode_function(self, generator='lambdify', **kwargs):
 
         """
 
-        if 'specified' in kwargs:
-            kwargs.pop('specified')
-            print("User supplied 'specified' kwarg was disregarded.")
+        try:
```
tarzzz: Why are you putting everything in a `try-except` block? You can simply check if "Specified" is in kwargs, then issue a warning, and move ahead.

jinankjain: As Jason mentioned he wants support of old library still so I used to try-except block

moorepants: Yes, please make this backwards compatible. There is a lot of code in the wild that would like to get a warning about this minor change.

jinankjain: Yes, this one supports both versions as I have imported both function.

moorepants:
I just read the code and realized that all this is not necessary. System handles `specifieds` internally. So I think all that needs to happen is to change that. The user doesn't ever type `specifieds` and if they do pass it to the `kwargs` it gets popped out. Deprecation is unlikely needed.

moorepants:
This function needs to be changed to use `specifieds`: `_kwargs_for_gen_ode_func()`

moorepants:
And `generate_ode_function` needs to pop either `specifieds` or `specified` because the user could pass in either.

jinankjain:
I have incorporated all the changes could you please review it.

moorepants:
@jinankjain I think it would help if you wrote a test first. For example, `System()` should be able to accept `specifieds` or `specified` as a key word arg.

moorepants:
Also, shouldn't we always be using the new `generate_ode_function`? Why use the old one at all? Isn't the whole purpose to remove the old version?

moorepants:
The new `generate_ode_function` only accepts `specifieds`.

You need to write a unit test for this first. If you write a unit test(s) for the `__init__` method then you can run the tests locally to see if they pass.

moorepants:
@jinankjain I've added a checklist to the top of the PR. Once you address those issues and a reviewer has checked them off, we can merge.

moorepants:
PR #154 fixes the issue. Closing this one.

Issue #138, opened 2015-03-05
kane._q and kane._u changed from lists to matrices in SymPy 0.7.6


coolyashish:
Hi I tried to create visualization for three_link_canical_pendulums but I couldn't due to type error. 

Commands to produce error:
1. Go to /pydy/examples/three_link_conical_pendulum
2. python simulate.py ..
3. python visualize.py
TypeError...

Error Logs: https://gist.github.com/coolyashish/c48909e428b3e05c94a9

moorepants:
@tarzzz You have any idea?

tarzzz:
@moorepants I have been closely looking at this particular issue. 
We have 31 arguments, i.e. 6 kane._q's, 6 kane._u's and 19 parameters(lengths, masses and inerties etc.). I fail to understand why is the numeric transform function requires only 25. Somehow it is neglecting some of parameters. Still looking into it.

tarzzz:
Ok, Here is my observation. When we do `kane._q + kane._u`, we get this:

    Matrix([[alpha0(t) + omega0(t)], [alpha1(t) + omega1(t)], [alpha2(t) + omega2(t)], [beta0(t) + delta0(t)], [beta1(t) + delta1(t)], [beta2(t) + delta2(t)]])

Instead of what should have been:

    Matrix([alpha0(t),alpha1(t), alpha2(t), beta0(t), beta1(t), beta2(t), omega0(t), omega1(t), omega2(t), delta0(t), delta1(t), delta2(t)])

This is because, a print on kane._q, and kane._u gives:

    Matrix([[alpha0(t)], [alpha1(t)], [alpha2(t)], [beta0(t)], [beta1(t)], [beta2(t)]])

and 
    
    Matrix([[omega0(t)], [omega1(t)], [omega2(t)], [delta0(t)], [delta1(t)], [delta2(t)]])

It is a list of lists, which is *probably* causing the issue. @moorepants Could you please confirm if any recent change in KanesMethod might have caused it? We can do a quick fix here, but instead we should look into the buggy code!

tarzzz:
@coolyashish : As a quick fix, you can do this:

In the example's `visualize.py` replace:

    scene.generate_visualization_json(kane._q + kane._u, param_syms,
                                  state_trajectories, param_vals)

with:

    l = []
    l+= [i for i in kane._q]
    l+= [i for i in kane._u]
    scene.generate_visualization_json(l, param_syms,
                                  state_trajectories, param_vals)

[...]

PR #161, opened 2015-05-10
Remove instances of _q and _u from examples.

moorepants:
Fixes issue #138.

- Minor edit to the three link pendulum.
- Removes mass_spring example, as it was an early hacked demo and does not use
  the current API, plus the mass_spring_damper is a better example that does
  everything it did.

  commit message.


moorepants:
@pydy/pydy-developers This PR needs a review, thanks!

sahilshekhawat:
looks like there are merge conflicts, can you please fix them?

moorepants:
Merge conflicts fixed.

examples/three_link_conical_pendulum/simulate.py
```diff
@@ -6,7 +6,7 @@
 from pydy.codegen.code import generate_ode_function
```
sahilshekhawat: Shouldn't we migrate our examples to new ``generate_ode_function``?

sahilshekhawat:
These changes does fix ``TypeError: <lambda>() takes exactly 25 arguments (31 given)`` which occurs in master but animation still doesn't work. Is it a known issue?

examples/three_link_conical_pendulum/simulate.py
moorepants: I can fix that one here. Should be a quick fix.

moorepants:
The animation browser works for me. This example works perfectly on my machine from derivation to visualization. Can you specify what isn't working for you?

moorepants:
> These changes does fix TypeError: <lambda>() takes exactly 25 arguments (31 given) which comes in master

Yes this is the error this is trying to fix.

sahilshekhawat:
It is working fine now, animation was not working before. Thanks for this fix :)

PR #165, opened 2015-05-11
Fixed broken benchmarks and made it work with new models

### LINKED ISSUE #164: Benchmarks are broken in master
There is an UnboundLocalError when I ran benchmarks
```
Pendulum with 1 links.
======================
The derivation took 0.12791 seconds.

  File "bin/benchmark_pydy_code_gen.py", line 118, in <module>
    run_benchmark(args.max_num_links, args.num_time_steps)
  File "bin/benchmark_pydy_code_gen.py", line 49, in run_benchmark
    x0 = hstack((0, pi / 2 * ones(len(results[3]) - 1), 1e-3 *
UnboundLocalError: local variable 'results' referenced before assignment
```
I think while we fix this, we should remove it from README.

sahilshekhawat:
Fixes #164

  commit message.
  in the unit test.)
  follow deprecation cycles.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,4 @@ nosetests.xml
 .mr.developer.cfg
 .project
 .pydevproject
+.idea/
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,5 +54,3 @@ script:
   - cd docs && make html && cd ..
   # Make sure the library installs.
   - python setup.py install
-  # Make sure benchmarks runs
-  - python bin/benchmark_pydy_code_gen.py 2 4
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,3 +54,5 @@ script:
   - cd docs && make html && cd ..
   # Make sure the library installs.
   - python setup.py install
+  # Make sure benchmarks runs
+  - python bin/benchmark_pydy_code_gen.py 2 4
--- a/bin/benchmark_pydy_code_gen.py
+++ b/bin/benchmark_pydy_code_gen.py
@@ -48,9 +48,7 @@ def run_benchmark(max_num_links, num_time_steps=1000):
 
         times = linspace(0, 10, num_time_steps)
         sys.times = times
-        print sys.coordinates
-        print sys.speeds
-        print sys.states
+
         x0 = hstack(
             (0,
              pi / 2 * ones(len(sys.coordinates) - 1),
--- a/bin/benchmark_pydy_code_gen.py
+++ b/bin/benchmark_pydy_code_gen.py
@@ -4,9 +4,10 @@
 import time
 
 # external libraries
-from numpy import linspace, zeros, zeros_like
+from numpy import hstack, ones, pi, linspace, array, zeros, zeros_like
 import matplotlib.pyplot as plt
 from pydy.models import n_link_pendulum_on_cart
+from sympy import symbols
 
 
 def run_benchmark(max_num_links, num_time_steps=1000):
@@ -30,6 +31,11 @@ def run_benchmark(max_num_links, num_time_steps=1000):
 
         start = time.time()
         sys = n_link_pendulum_on_cart(n, cart_force=False)
+
+        m = symbols('m:{}'.format(n + 1))
+        l = symbols('l:{}'.format(n))
+        g = symbols('g')
+
         derivation_times[j] = time.time() - start
         print('The derivation took {:1.5f} seconds.\n'.format(derivation_times[j]))
 
@@ -42,6 +48,20 @@ def run_benchmark(max_num_links, num_time_steps=1000):
 
         times = linspace(0, 10, num_time_steps)
         sys.times = times
+        print sys.coordinates
+        print sys.speeds
+        print sys.states
+        x0 = hstack(
+            (0,
+             pi / 2 * ones(len(sys.coordinates) - 1),
+             1e-3 * ones(len(sys.speeds))))
+        sys.initial_conditions = dict(zip(sys.states, x0))
+
+        constants = [g, m[0]]
+        for i in range(n):
+            constants += [l[i], m[i + 1]]
+
+        sys.constants = dict(zip(constants, array(parameter_vals)))
 
         for k, method in enumerate(methods):
 
--- a/bin/benchmark_pydy_code_gen.py
+++ b/bin/benchmark_pydy_code_gen.py
@@ -2,13 +2,9 @@
 
 # standard library
 import time
-import os
-import shutil
-import glob
 
 # external libraries
-from scipy.integrate import odeint
-from numpy import hstack, ones, pi, linspace, array, zeros, zeros_like
+from numpy import linspace, zeros, zeros_like
 import matplotlib.pyplot as plt
 from pydy.models import n_link_pendulum_on_cart
 
@@ -44,11 +40,8 @@ def run_benchmark(max_num_links, num_time_steps=1000):
         for i in range(n):
             parameter_vals += [arm_length, bob_mass]
 
-        # odeint arguments
-        x0 = hstack((0, pi / 2 * ones(len(sys.coordinates) - 1), 1e-3 *
-                    ones(len(sys.speeds))))
-        args = {'constants': dict(zip(sys.constants, array(parameter_vals)))}
-        t = linspace(0, 10, num_time_steps)
+        times = linspace(0, 10, num_time_steps)
+        sys.times = times
 
         for k, method in enumerate(methods):
 
@@ -56,23 +49,18 @@ def run_benchmark(max_num_links, num_time_steps=1000):
             print(subtitle)
             print('-' * len(subtitle))
             start = time.time()
-            evaluate_ode = sys.generate_ode_function(generator=method)
+            sys.generate_ode_function(generator=method)
             code_generation_times[j, k] = time.time() - start
             print('The code generation took {:1.5f} seconds.'.format(
                 code_generation_times[j, k]))
 
             start = time.time()
-            odeint(evaluate_ode, x0, t, args=(args,))
+            sys.integrate()
             integration_times[j, k] = time.time() - start
             print('ODE integration took {:1.5f} seconds.\n'.format(
                 integration_times[j, k]))
 
-        del sys, evaluate_ode
-
-    # clean up the cython crud
-    files = glob.glob('multibody_system*')
-    for f in files:
-        os.remove(f)
+        del sys
 
     # plot the results
     fig, ax = plt.subplots(3, 1, sharex=True)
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -48,3 +48,4 @@ nosetests.xml
 .mr.developer.cfg
 .project
 .pydevproject
+.idea/
```
moorepants: What is this for?

bin/benchmark_pydy_code_gen.py
```diff
@@ -69,13 +67,12 @@ def run_benchmark(max_num_links, num_time_steps=1000):
             print('ODE integration took {:1.5f} seconds.\n'.format(
                 integration_times[j, k]))
 
-        del results, evaluate_ode
+        del sys, evaluate_ode
 
     # clean up the cython crud
```
moorepants: All of this can go as the Cython generator no longer creates these directories.

moorepants:
@sahilshekhawat Can you please add a simple run of this to the `.travis.yml` file? That way we will catch if this breaks in the future. You could add a new item to the matrix for examples and benchmarks. We will eventually automate the testing of the examples too.

moorepants:
Got this error when I ran it:

```
/home/moorepants/src/pydy/pydy/codegen/code.py:18: DeprecationWarning: This module, 'pydy.codgen.code', is deprecated. The function 'generate_ode_function' can be found in the 'pydy.codegen.ode_function_generator' module. 'CythonGenerator' has been removed, use 'pydy.codegen.cython_code.CythonMatrixGenerator' instead.
  DeprecationWarning)
Pendulum with 1 links.
======================
The derivation took 0.09895 seconds.

Generating with lambdify method.
--------------------------------
/home/moorepants/src/pydy/pydy/codegen/code.py:68: DeprecationWarning: This function is deprecated and will be removed in PyDy 0.4.0. Use the the new 'generate_ode_function' in 'pydy.codegen.ode_function_generator'
  DeprecationWarning)
The code generation took 0.00751 seconds.
  File "bin/benchmark_pydy_code_gen.py", line 115, in <module>
    run_benchmark(args.max_num_links, args.num_time_steps)
  File "bin/benchmark_pydy_code_gen.py", line 65, in run_benchmark
    odeint(evaluate_ode, x0, t, args=(args,))
  File "/home/moorepants/anaconda/envs/pydy-dev-examples/lib/python2.7/site-packages/scipy/integrate/odepack.py", line 148, in odeint
    ixpr, mxstep, mxhnil, mxordn, mxords)
  File "/home/moorepants/src/pydy/pydy/codegen/ode_function_generators.py", line 511, in rhs
    args = self._parse_all_args(*args)
  File "/home/moorepants/src/pydy/pydy/codegen/ode_function_generators.py", line 446, in _parse_all_args
    args = self._parse_constants(*args)
  File "/home/moorepants/src/pydy/pydy/codegen/ode_function_generators.py", line 399, in _parse_constants
    p = self._convert_constants_dict_to_array(p)
  File "/home/moorepants/src/pydy/pydy/codegen/ode_function_generators.py", line 388, in _convert_constants_dict_to_array
    self._constants_values[i] = p[c]
KeyError: l0
```

@sahilshekhawat Can you please review one or more of my PRs and I'll review this one for you?

bin/benchmark_pydy_code_gen.py
```diff
@@ -46,9 +45,9 @@ def run_benchmark(max_num_links, num_time_steps=1000):
             parameter_vals += [arm_length, bob_mass]
 
         # odeint arguments
-        x0 = hstack((0, pi / 2 * ones(len(results[3]) - 1), 1e-3 *
-                    ones(len(results[4]))))
-        args = {'constants': dict(zip(results[2], array(parameter_vals)))}
+        x0 = hstack((0, pi / 2 * ones(len(sys.coordinates) - 1), 1e-3 *
+                    ones(len(sys.speeds))))
+        args = {'constants': dict(zip(sys.constants, array(parameter_vals)))}
```
moorepants: I think this needs to be `sys.constants_symbols`.

moorepants:
The `odeint` call should be changed to `System.integrate()`.

bin/benchmark_pydy_code_gen.py
moorepants: I also don't think this matches the order of the constant symbols.

bin/benchmark_pydy_code_gen.py
```diff
@@ -46,9 +45,9 @@ def run_benchmark(max_num_links, num_time_steps=1000):
             parameter_vals += [arm_length, bob_mass]
 
         # odeint arguments
-        x0 = hstack((0, pi / 2 * ones(len(results[3]) - 1), 1e-3 *
-                    ones(len(results[4]))))
-        args = {'constants': dict(zip(results[2], array(parameter_vals)))}
+        x0 = hstack((0, pi / 2 * ones(len(sys.coordinates) - 1), 1e-3 *
```
moorepants: This should be changed to a dictionary and set with `sys.initial_conditions=`.

.gitignore
sahilshekhawat: This is for PyCharm.

moorepants: Ok, that's fine.

sahilshekhawat:
@moorepants Why do we need to add another item to ``matrix``? We can just add the commands to the ``script``. I mean do we need a different environment for examples?

moorepants:
I suggested a new matrix so that we only run the examples once on testing. If you add it to script then it runs on every matrix item, which is not necessary. But it is fine if you add it to script for now since it is just one small thing.

sahilshekhawat:
@moorepants benchmarks are working but I am getting ``AttributeError: 'module' object has no attribute 'packaging'`` in travis when I run ``benchmark_pydy_code_gen.py 2 4``.  Can you please see if you can help me?

.travis.yml
```diff
@@ -54,3 +56,5 @@ script:
   - cd docs && make html && cd ..
   # Make sure the library installs.
   - python setup.py install
+    # Make sure benchmarks runs
+  - benchmark_pydy_code_gen.py 2 4
```
moorepants: add a newline at the end

moorepants: Don't you need to use `python bin/benchmark_pydy_code_gen.py 2 4`?

moorepants: Also note that matlplotlib is not installed on Travis. Either we need to install it or make the plot optional in the benchmark script. I think I'd prefer the latter.

.travis.yml
```diff
@@ -37,6 +37,8 @@ install:
   - pip install sphinx numpydoc
   # Install PEP 8 linters (flake8 is like pep8 but also reports unused imports)
   - pip install flake8
+  # Install matplotlib for benchmarks
+  - pip install matplotlib
```
moorepants: Installing matploblib with pip is generally a bad idea. If you want to install it, install it with apt-get.

moorepants:
I'm not completely sure what is causing the Travis error.

moorepants:
Maybe we need a `pip --upgrade setuptools pip` line to make sure those are up-to-date. Maybe it is using the system setuptools and it is old or something.

.travis.yml
sahilshekhawat: Since pydy is installed before this runs, it works. I have also tried ``python bin/benchmark_pydy_code_gen.py 2 4``. Okay I will make plot optional.

.travis.yml
sahilshekhawat: I am leaving the plot in travis. thank you for the advice, will use apt-get from now.

.travis.yml
```diff
@@ -54,3 +56,7 @@ script:
   - cd docs && make html && cd ..
   # Make sure the library installs.
   - python setup.py install
+  # Switch back to latest setuptools
+  - pip --upgrade setuptools pip
```
moorepants: I'd move this to the beginning of all the installs. Maybe right after: `- sudo apt-get update`

.travis.yml
```diff
@@ -37,6 +37,8 @@ install:
   - pip install sphinx numpydoc
   # Install PEP 8 linters (flake8 is like pep8 but also reports unused imports)
   - pip install flake8
+  # Install matplotlib for benchmarks
+  - sudo apt-get install python-matplotlib
```
moorepants: Combine this with the other apt-get installs above.

moorepants:
I can't find anything useful about this error. We can skip the Travis testing of this file if you want and just open an issue. It isn't critical for the next release.

sahilshekhawat:
.I have opened an issue, leaving adding benchmark's run to travis for now.

.travis.yml
```diff
@@ -18,9 +18,6 @@ install:
   - sudo apt-get install -y -qq python-numpy python-scipy cython
   - sudo apt-get install -y -qq python-coverage
   - sudo apt-get install -y -qq phantomjs
-  # SymPy's development version identifiers are not PEP 440 compliant and
```
moorepants: These changes seem like the ones that I committed in my last PR.

moorepants:
This is good to merge except that I don't understand the diff for the .travis.yml file.

sahilshekhawat:
Removed extra commits and reverted ``.travis.yml``

bin/benchmark_pydy_code_gen.py
```diff
@@ -45,11 +40,8 @@ def run_benchmark(max_num_links, num_time_steps=1000):
         for i in range(n):
             parameter_vals += [arm_length, bob_mass]
 
-        # odeint arguments
-        x0 = hstack((0, pi / 2 * ones(len(results[3]) - 1), 1e-3 *
-                    ones(len(results[4]))))
-        args = {'constants': dict(zip(results[2], array(parameter_vals)))}
```
moorepants: I just noticed this. Sorry.

`x0` needs to be set in system as `sys.initial_conditions` and `constants` needs to be set as `sys.constants`.

moorepants:
@sahilshekhawat What is the status here? Can you fix the errors I noted above?

bin/benchmark_pydy_code_gen.py
sahilshekhawat: I want to ask one thing, here x0  is an array but while setting initial conditions of a system it checks if its a dict and whether it matches its states to floats i.e. initial_conditions.keys(). Do both of them works with odeint? I have used zip() with x0 and sys.states to create a dict, is it correct?

moorepants: The System class only accepts a dict for the initial conditions, so yes you can zip() x0 and the states (coordinates and speeds) to create that dict.

moorepants: See: http://pydy.readthedocs.org/en/latest/system.html#pydy.system.System

sahilshekhawat: Thanks, you can review it whenever you have time.

moorepants:
Thanks @sahilshekhawat, looks great!

PR #168, opened 2015-05-20
Mocked out more modules for readthedocs.

moorepants:


moorepants:
I'm just going to merge this into master to see if the RTD builds. This isn't following protocol but this is the easiest way to test and there is little it can hurt.

PR #169, opened 2015-05-20
Added one more mocked module and dealt with sympy verion checks on mo…

moorepants:
…cked module.


moorepants:
Also merging for RTD testing.

Issue #89, opened 2014-08-14
Update the release notes with all pydy.viz changes


moorepants:

moorepants:
@tarzzz The release notes for 0.3.0 still need to be updated to reflect all the stuff from your GSoC work last summer.

PR #177, opened 2015-05-22
Updated the release notes with visualizer changes.

moorepants:
Fixes issue #89.
Also fixes some spelling errors.

  commit message.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -81,8 +81,8 @@ Anaconda_ distribution for easy cross platform installation.
 .. _conda: http://conda.pydata.org/
 .. _Anaconda: http://docs.continuum.io/anaconda/
 
-Once the dependencies are installed, the lastest stable version of the package
-can be downloaed from PyPi::
+Once the dependencies are installed, the latest stable version of the package
+can be downloaded from PyPi::
 
    $ wget https://pypi.python.org/packages/source/p/pydy/pydy-0.2.1.tar.gz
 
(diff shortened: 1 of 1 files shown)


moorepants:
@pydy/pydy-developers This needs review and merging. Simple README changes.

README.rst
```diff
@@ -343,11 +343,29 @@ User Facing
...
+  - The visualizer can now be loaded in an IPython notebook via IPython's
+    widgets.
+  - A slider was added to manually control the frame playback.
+  - The visualization objects can be manipulated via the GUI.
+  - The scene json file can be edited via the GUI.
+  - The data and scene description have been separated into two json files.
+  - Different simulation files can be loaded via the GUI.
+  - Shapes can now have a material.
+  - Simulation parameters can be modified and the simulations can be rerun all
+    via the GUI.
```
jcrist: Does all of this actually work now? If not, we should probably wait to add it to the release notes.

moorepants: Yes, it essentially works. But even so, all of these changes are merged into master already, so the release notes should reflect what is in master.

I'm starting to work on the viz bugs associated with these changes now. But I don't see any reason not to merge this pr.

jcrist: As long as these are actually implemented and working when we release, then yeah, everything's fine. I'd just hate to say "The visualizer can now be loaded in an IPython notebook via IPython's widgets", and have that not actually be true when we release. Not a -1, just a note. Looks good.

moorepants: Sure, if any of these things are not true come time for a release it needs to be modified. Right now ipython notebooks work except that specifying cameras no longer works.

moorepants:
If you think this is good, check off the checklist and merge.

PR #183, opened 2015-05-26
Added window resize event listener to the canvas

### LINKED ISSUE #35: UI doesn't look good in small window sizes
_Issue by **[moorepants](https://github.com/moorepants)** from Sunday Feb 23, 2014 at 05:27 GMT_
_Originally opened as https://github.com/pydy/pydy-viz/issues/112_

----

screenshot from 2014-02-22 17 08 52

### LINKED ISSUE #202: The "Play Looped" checkbox should be beside the play buttons and slider
Right now it is placed on the right side away from the play controls. This should be near the play controls.

sahilshekhawat:
Fixes #35
Fixes #202

  commit message.
  in the unit test.)
  follow deprecation cycles.)


sahilshekhawat:
@moorepants I was wondering if we can increase the width of the visualizer to use the extra space on the right side?

moorepants:
Nice fix! Here are some comments, they may or may not need be addressed here.

> @moorepants I was wondering if we can increase the width of the visualizer to use the extra space on the right side?

Absolutely, it seems like there is a big gap of wasted space. Here is what it looks like on a big screen:
pydy visualizer - mozilla firefox_081

Then if I close in the width the canvas changes accordingly, but notice that the buttons and slider end up covering the input text box and such:
pydy visualizer - mozilla firefox_082

The links seem to get thinner too. We do not want the scale of the objects in the window to change. The aspect ratio should remain the same in the canvas window, i.e. the ratio of width to height of the objects.

And is there any reason we should have the height squeeze the window too:
pydy visualizer - mozilla firefox_083

I'm thinking maybe we should.

Lastly: how does this change affect the IPython Notebook widget?

sahilshekhawat:
I didn't looked into buttons and sliders but will fix that. About height, I don't know but its fixed to 480px and that's why links gets thinner, I am updating it to be square.
I have not tested that but after seeing the implementation I should work just fine.

moorepants:
Maybe we should set a minimum square canvas size. If the window is larger then it can fill the space as a rectangle if the window is too small then we add scroll bars. That's an idea.

Or if you can scale things properly and make it very responsive that is fine too. But the objects have to have equal aspect ratio in any window size.

pydy/viz/static/js/dyviz/scene.js
```diff
@@ -95,6 +95,34 @@ DynamicsVisualizer.Scene = Object.extend(DynamicsVisualizer, {
 
     },
 
+    WindowResize: function(renderer, camera){
+        /**
+          * Adds window resize event listener
+          * to renderer and camera and updates
+          * them accordingly. This is a modified
+          * version of THREEX.WindowResize.js
```
moorepants: Please note the license that this is shared under here.

sahilshekhawat: Its MIT LIcense, Okay I will mention it too.

sahilshekhawat:
I was thinking that instead of showing editable file name in load simulation input field maybe we should give an option of browse files. what you guys think?

moorepants:
See this issue https://github.com/pydy/pydy/issues/189, let's do that in a separate PR.

moorepants:
Does this PR need review? What is the current state?

sahilshekhawat:
I am currently working on it. No offence to anyone but the code is very bad. for example this is the canvas tag ``<canvas width="918" height="290" style="width: 1020.68px; height: 322.4px;"></canvas>``

moorepants:
@sahilshekhawat There is no reason to say that the code is bad. Please refrain from doing that in the future. We are only interested in making it better. In that sense, all code is bad, improving is a continual process. If you think you can improve something, just do it. Code is what it is. There are always deadlines and other constraints that make the code what it is.

sahilshekhawat:
@moorepants I apologize, I shouldn't have said that. I just wanted to let tell you why it is taking soo much time. I will take care of this from now on.

moorepants:
NP, I'm not worried that it is taking more time than expected. Everything always takes more time than expected!

sahilshekhawat:
@moorepants Its ready for review.

moorepants:
I just tried out the three link conical pendulum example (not in a notebook). I like it! So you have a minimum canvas window size and it will fill the space if you have a big resolution and if you don't it will show the minimum size. The aspect ratio stays fixed during resizing which is correct too. The buttons for play, pause, stop are always visible and the play looped checkbox is near the buttons, all good.

I noticed one odd behavior:

If you click the blue slider it jumps to the right. If you click and drag it jumps to the right before dragging and it follows your mouse but it doesn't track your mouse (i.e. stay directly under the mouse). I don't recall this happening before. Maybe something was slightly changed that affected this.

pydy/viz/static/index.html
```diff
@@ -96,13 +89,13 @@
               <span class="glyphicon glyphicon-edit" ></span>
               Show Scene JSON
             </button>
+            <label id="checkbox-play-looped"><input type="checkbox" id="play-looped">Play Looped</label>
             <div id="slider-wrapper">
               <span class="glyphicon glyphicon-time" ></span>
-              <span style="margin-right:1em;" id="time"> 0s</span>
-              <div id="time-slider" class="slider" style="float:right; width:28em;padding-left: 2em;"></div>
+              <span style="margin-right:1em;" id="time"> 0.000 s</span>
```
moorepants: Humans can only detect up to 60 hz (~0.02 s), so I don't think showing a third decimal here is needed. Showing only two digits after the decimal should always be sufficient.

But saying that, we currently animate at whatever framerate that is dictated by the time array that is generated. There is a bigger issue here about whether we want to animate at very slowed down frame rates or not.

Anyways, for now we can just stick to two digits if that is  quick fix.

moorepants:
I tried out the mass_spring_damper example (in a notebook) and see this:
selection_090

We should probably maximize the width of the canvas in the notebook. I suggest moving shape editor dialog up beside the parameter input box so that we can have the full width for the canvas.

Note that the pause/stop buttons are not present, nor is the "play looped" button.

sahilshekhawat:
Yes, thanks for reminding me. I totally forgot to do the same changes for ipython too. Fixing.

moorepants:
Note that the slider follows the mouse correctly in the notebook, so there is something that was introduced here that is goofing up the slider. Not sure what it is.

sahilshekhawat:
>We should probably maximize the width of the canvas in the notebook. I suggest moving shape editor dialog up beside the parameter input box so that we can have the full width for the canvas.
I can't move it up with paramter input boxes but I can put it over canvas, just two buttons of edit objects and close dialog.

moorepants:
Whatever you think looks good.

sahilshekhawat:
Ready for review. Fixed Ipython visualization UI issues.

moorepants:
The slider is crammed into the space beside "Play Looped" for me:
selection_091

moorepants:
@sahilshekhawat Let's try to get this merged without adding new things to it. Add new things in new PRs.

sahilshekhawat:
pydy-ipython
For me its different.
Okay lets get this merged first.

moorepants:
Is it because I may have a wider screen?

sahilshekhawat:
I don't think it should be the reason but i don't know, its working same on firefox and chrome. so browser can't be the issue.

sahilshekhawat:
Oh! yes, I got that and yes its because you have wider screen.

Issue #182, opened 2015-05-26
Visualization examples do not work sometimes


sahilshekhawat:
This bug is consistence in examples like double pendulum,  three link conical pendulum etc. Sometimes examples do not work and I am yet to find the reason (but they mostly happen when there is no file from previous run). Instead of the visualization of the example they shows the visualization as in the image below
![part1](https://cloud.githubusercontent.com/assets/5890452/7806484/13a6cbaa-039f-11e5-9cb2-649148f94db4.png)

sahilshekhawat:
The correct visualization is shown on reload.

sahilshekhawat:
Reloading after sometime works though!

sahilshekhawat:
I think this is responsible [L23](https://github.com/pydy/pydy/blob/master/pydy/viz/static/js/dyviz/parser.js#L23)
As, the call is not completed before Scene and thus, json file is not loaded.

moorepants:
@sahilshekhawat Do you still see this problem? I haven't encountered it in a while.

sahilshekhawat:
Yeah! I just tried it and it works right away.

moorepants:
Ok, let's close this. We can reopen it if we come across this again.

PR #185, opened 2015-05-26
Fixed several visualizer bugs.

### LINKED ISSUE #85: [Viz: GUI] Unable to change color when material is default
When the material is set to Default in GUI, Color is not shown on the objects, while it works fine when other material(Other than default) is chosen

### LINKED ISSUE #94: Trackball controls broken, when user defined camera is added
When A custom camera is added from Python side to Scene, the trackball controls do not work

### LINKED ISSUE #96: Colors do not work in GUI
Colors specified to objects do not work properly in GUI. Sometimes they are shown, sometimes only grey color appears. Sometimes changing color of one object changes the colors of all other objects as well.

### LINKED ISSUE #151: Cameras Not working in Master
The `addCamera` method in [L126](https://github.com/pydy/pydy/blob/master/pydy/viz/static/js/dyviz/scene.js#L126) messes up with the `_addDefaultCamera` method in [L48](https://github.com/pydy/pydy/blob/master/pydy/viz/static/js/dyviz/scene.js#L48). This leads to Cameras freezing once the animation starts/objects are loaded. As a temporary workaround, I have removed the addCamera functionality. As a result, the cameras passed into JSON object are not added to the scene.

### LINKED ISSUE #174: The colors are not correct in the three link pendulum example.
pydy visualizer - mozilla firefox_080

### LINKED ISSUE #187: Editing materials via the GUI causes the camera to freeze and the object size to change
To reproduce:

1. Run `python visualize.py` in the `three_link_conical_pendulum` example.
2. Open link in browser.
3. Click "Edit Objects" and select the first cylinder.
4. Change the material to an option other than default and click "Apply".

The object changes to the material, the object size grows, and the trackballcontrols lock.

sahilshekhawat:
Fixes #85
Fixes #94 
Fixes #96
Fixes #151
Fixes #174 
Fixes #187

  commit message.
  in the unit test.)
  follow deprecation cycles.)

### DIFF
--- a/pydy/viz/static/js/dyviz/param_editor.js
+++ b/pydy/viz/static/js/dyviz/param_editor.js
@@ -124,7 +124,6 @@ DynamicsVisualizer.ParamEditor = Object.extend(DynamicsVisualizer, {
 
         jQuery.extend(true,self.model.objects[int_id],updated_object);
         self.Scene.addObjects();
-        //self.Scene.addCameras();
         self.Scene.addLights();
         self.loadUIElements();
     },
--- a/pydy/viz/static/js/dyviz/param_editor.js
+++ b/pydy/viz/static/js/dyviz/param_editor.js
@@ -26,14 +26,8 @@ DynamicsVisualizer.ParamEditor = Object.extend(DynamicsVisualizer, {
         div1.append(jQuery('<input />',{ type:'text', id: "_color", class: 'form-control', value: toLoad.color}));
 
         var div_material = jQuery('<select />',{class: 'form-control', id:"_material"});
-        for(var i in self.Materials){
-            if(i == toLoad.type) {
-                div_material.append('<option value="' + i +  '" selected="selected" >' + i + '</option>');
-
-            }
-            else {
-                div_material.append('<option value="' + i +  '">' + i + '</option>');
-            }
+        for(var i=0;i<self.MaterialsList.length; i++){
+            div_material.append('<option value="' + self.MaterialsList[i] +  '">' + self.MaterialsList[i] + '</option>');
         }
 
         var div_geom = jQuery('<select />',{class: 'form-control', id:"_geometry"});
--- a/pydy/viz/static/js/dyviz/param_editor.js
+++ b/pydy/viz/static/js/dyviz/param_editor.js
@@ -27,7 +27,7 @@ DynamicsVisualizer.ParamEditor = Object.extend(DynamicsVisualizer, {
 
         var div_material = jQuery('<select />',{class: 'form-control', id:"_material"});
         for(var i=0;i<self.MaterialsList.length; i++){
-            if(self.Geometries[i] == toLoad.type){
+            if(self.MaterialsList[i] == toLoad.type){
                 div_material.append('<option value="' + self.MaterialsList[i] +  '" selected="selected">' + self.MaterialsList[i] + '</option>');
             }  else {
                 div_material.append('<option value="' + self.MaterialsList[i] +  '">' + self.MaterialsList[i] + '</option>');
--- a/pydy/viz/static/js/dyviz/param_editor.js
+++ b/pydy/viz/static/js/dyviz/param_editor.js
@@ -27,7 +27,11 @@ DynamicsVisualizer.ParamEditor = Object.extend(DynamicsVisualizer, {
 
         var div_material = jQuery('<select />',{class: 'form-control', id:"_material"});
         for(var i=0;i<self.MaterialsList.length; i++){
-            div_material.append('<option value="' + self.MaterialsList[i] +  '">' + self.MaterialsList[i] + '</option>');
+            if(self.Geometries[i] == toLoad.type){
+                div_material.append('<option value="' + self.MaterialsList[i] +  '" selected="selected">' + self.MaterialsList[i] + '</option>');
+            }  else {
+                div_material.append('<option value="' + self.MaterialsList[i] +  '">' + self.MaterialsList[i] + '</option>');
+            }
         }
 
         var div_geom = jQuery('<select />',{class: 'form-control', id:"_geometry"});
@@ -120,7 +124,7 @@ DynamicsVisualizer.ParamEditor = Object.extend(DynamicsVisualizer, {
 
         jQuery.extend(true,self.model.objects[int_id],updated_object);
         self.Scene.addObjects();
-        self.Scene.addCameras();
+        //self.Scene.addCameras();
         self.Scene.addLights();
         self.loadUIElements();
     },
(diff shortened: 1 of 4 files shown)


moorepants:
I checked out this PR, ran the three link pendulum visualization (manually removed static folder first) and no longer see any objects in the canvas. The edit objects dialog box is empty and no scene.json seems to be loaded. This is issue #182. When I reload the browser the objects load properly and the colors are now correct but when I try to edit the colors in the "Edit Objects" dialog box the colors do no update, instead objects disappear from view.

moorepants:
Also the "Edit Objects" "material" dropdown is filled with various methods instead of a list of materials.

sahilshekhawat:
Yes,  it's exactly like you explained.  Both of those issues are there and i will be fixing them as well.Can I raise another pr for these two bug fixes? 
Fixing edit objects material drop down here itself.

moorepants:
On master, the "Edit Objects" "Material" dropdown has the correct materials listed. So this PR breaks that. We try our best to keep the master branch in an unbroken state* and this would break things if we merge as is. The fix for issue #182 can be in another PR.

*master is unfortunately in a broken state due to premature merging of the new visualizer, but we'd like to avoid making it more broken in different ways.

sahilshekhawat:
I apologize if I wasn't clear but I intended to fix that one here itself.  I was talking about issue #182 and colors are not changed from GUI,  both of which are broken in master and are known and reported bugs. 
But let's not go into that,  we both want to fix things. :smile:
Thanks

moorepants:
This PR title says it fixes colors yet it changes material behavior. It should stick to one, the other, or both. Right now it seems to fix the colors on initial load but breaks materials selection.

moorepants:
Couple more issues:

1. When the animation is playing I'm not able to control the zoom or camera rotation on the canvas.
2. The slider starts off on the right side of the "Play" button and when play is pressed it moves underneath, causing the whole canvas to move. My opinion is that the play and stop buttons should always be shown. The stop button should grey out when the animation is not playing.

sahilshekhawat:
That is what i wanted to discuss, the issue is that user defined camera is there in the simulation.json and it is being used during animation. I wanted to know if its intended. If not then we need to fix API.

sahilshekhawat:
Just like other elements in the scene. It is changed as per simluation.json during animation.

moorepants:
The user defined camera should only be set on initially loading of a simulation. Otherwise it should be left to be whatever the user sets the camera to with the trackball controls.

moorepants:
Actually, to be more correct, the user defined camera should only be set when on the initial loading of a scene. If you load more simulations via the gui the camera should not change.

1. User defines a camera in the python code which generates a Scene.
2. The visualization scene is loaded with the default simulation. The camera should match the user's defined camera.
3. The user presses play and it starts from the view defined in 1.
4. Press stop.
5. The user rotates, zooms, and pans to pick a new view of the scene.
6. The user presses play and it starts from the view defined in 5.
7. While the animation is playing the user rotates, zooms, and pans.
8. The user presses stop. The view should stop on what was set with #7.
9. The user edits the objects via the gui and the viewing should still be the same as 7.

pydy/viz/static/js/dyviz/param_editor.js
```diff
@@ -26,14 +26,8 @@ DynamicsVisualizer.ParamEditor = Object.extend(DynamicsVisualizer, {
...
-        for(var i in self.Materials){
-            if(i == toLoad.type) {
-                div_material.append('<option value="' + i +  '" selected="selected" >' + i + '</option>');
-
-            }
-            else {
-                div_material.append('<option value="' + i +  '">' + i + '</option>');
-            }
+        for(var i=0;i<self.MaterialsList.length; i++){
+            div_material.append('<option value="' + self.MaterialsList[i] +  '">' + self.MaterialsList[i] + '</option>');
```
moorepants: Does this change ensure that the material defined in the Python code for a shape is selected on load?

sahilshekhawat: Oh! Right It does not. Thanks

sahilshekhawat:
@moorepants I have fixed the camera orientation. I will fix slider in #183

pydy/viz/static/js/dyviz/param_editor.js
```diff
@@ -126,7 +124,7 @@ DynamicsVisualizer.ParamEditor = Object.extend(DynamicsVisualizer, {
 
         jQuery.extend(true,self.model.objects[int_id],updated_object);
         self.Scene.addObjects();
-        self.Scene.addCameras();
+        //self.Scene.addCameras();
```
moorepants: Any reason this is commented out instead of deleted?

Also, do custom cameras that are defined in the Python code work now that this commented out?

moorepants:
It now works as expected for me with the three link conical pendulum example. Nice work.

pydy/viz/static/js/dyviz/param_editor.js
sahilshekhawat: Yes! It will help in future if we need to debug. Tarun's commented code helped me a lot finding my way.
Yes custom cameras now work, this part is run when we click "Apply" in editobjects. That's why we don't need to set camera here.

moorepants: I think this line should be deleted. I don't see what it offers being there.

sahilshekhawat: Okay

moorepants:
I think this is good to merge. The only thing is JS unit tests, which I don't even know how or if they work. Can you check and see if any of the JS tests are relevant here?

sahilshekhawat:
Sure

moorepants:
I just looked at the tests which are pretty sparse. I don't think these changes would affect anything but it may be worth trying to run them and see.

Do you want to try to add some tests that check for these errors?

I think we need to figure out the JS testing because we can avoid merging in bad code in the future if we have tests in place.

sahilshekhawat:
All the tests are passing but I want to add some more tests.

sahilshekhawat:
Most of the tests are not implemented yet. I think that I should raise another PR for it and merge this one?

moorepants:
I think that you should write tests that test things you fixed here with this PR. The PR should contain the code and its associated tests. That way we never merge code into master that is not tested.

sahilshekhawat:
I have worked on tests today, trying to figure out how things are happening. The main problem is the AJAX requests. Tests are failling without an error and I saw that Tarun also tried it but then commented out all those tests. It a big enough issue, I was working  on tests today, all day. Also, there are just 15 tests, thus, we need to build all of our testing framework. Should I continue to fix tests? Or should I came back to it and work on something else for now?

moorepants:
Is it possible to had a handful of tests that test what you fixed in this PR?

moorepants:
I'm on gitter if you want to discuss this.

moorepants:
We merging without JS tests because the testing suite is not in a function state and will need to be addressed separately first.

PR #194, opened 2015-05-29
Migrated Ipython widgets to IPython 3

### LINKED ISSUE #142: Notebook visualizations no longer work now that IPython 3.0 is released.
This is some of the problem:

http://ipython.org/ipython-doc/3/whatsnew/version3_widget_migration.html

This notebook should work:

http://nbviewer.ipython.org/gist/moorepants/aab1a1c6b49f51718a90

sahilshekhawat:
Fixes #142 

  commit message.
  in the unit test.)
  follow deprecation cycles.)

### DIFF
--- a/pydy/viz/scene.py
+++ b/pydy/viz/scene.py
@@ -480,20 +480,20 @@ class Scene(object):
     def display_ipython(self):
         """
         Method to display the visualization inside the
-        Ipython notebook. It is only supported by IPython
+        IPython notebook. It is only supported by IPython
         versions>=2.0.0
 
         """
 
         # Raise error whenever display_ipython() is called
-        # and Ipython is not installed or Ipython < '2.0.0'
+        # and IPython is not installed or IPython < '2.0.0'
         if IPython is None:
-            raise ImportError('Ipython is not installed but is required. ' +
-                              'Please install IPython >= 2.0.0 and try again')
+            raise ImportError('IPython is not installed but is required. ' +
+                              'Please installed IPython >= 2.0 and try again')
         elif ipython_less_than_3 is None:
-            raise ImportError('Ipython install is ' + Ipython.__version__ +
-                              ' but PyDy supports Ipython >= 2.0.0 ' +
-                              'Please update Ipython and try again')
+            raise ImportError('You have IPython ' + IPython.__version__ +
+                              ' installed but PyDy supports IPython >= 2.0.0' +
+                              'Please update IPython and try again')
 
         self.create_static_html(silent=True)
         self._widget_dict = OrderedDict()
--- a/pydy/viz/scene.py
+++ b/pydy/viz/scene.py
@@ -8,6 +8,7 @@ import distutils
 import distutils.dir_util
 import datetime
 from collections import OrderedDict
+from pkg_resources import parse_version
 # external
 from sympy.physics.mechanics import ReferenceFrame, Point
 
@@ -24,14 +25,21 @@ try:
     from IPython.html import widgets
     from IPython.display import display, Javascript
 
-    # Support for ipython 2 and ipython 3
-    if IPython.__version__[0] == '2':
-        ipython2 = True
-    else:
-        ipython2 = False
+    ipy_ver = parse_version(IPython.__version__)
+    if ipy_ver < parse_version('2.0'):
+        raise ImportWarning('PyDy supports IPython >= 2.0.0, while you have ' +
+                            'IPython ' + IPython.__version__ + ' installed. ' +
+                            'IPython related functionalities will not be ' +
+                            'available')
+        ipython_less_than_3 = None
+    elif ipy_ver >= parse_version('2.0') and ipy_ver < parse_version('3.0'):
+        ipython_less_than_3 = True
+    else:  # ipython >= 3.0
+        ipython_less_than_3 = False
 
 except ImportError:
     IPython = None
+    ipython_less_than_3 = None
 
 
 class Scene(object):
@@ -418,9 +426,10 @@ class Scene(object):
         simulation_outfile.write(json.dumps(self._simulation_data_dict, indent=4,
                                 separators=(',', ': ')))
         simulation_outfile.close()
-        if not silent: print("To view the visualization, open {}".format(
-                            os.path.join(dst, 'index.html')) +
-                            " in a WebGL compliant browser.")
+        if not silent:
+            print("To view the visualization, open {}".format(
+                  os.path.join(dst, 'index.html')) +
+                  " in a WebGL compliant browser.")
 
     def remove_static_html(self, force=False):
         """Removes the ``static`` directory from the current working
@@ -439,13 +448,13 @@ class Scene(object):
 
         if not force:
             ans = raw_input("Are you sure you would like to delete the " +
-                                "'static' directory? [y|n]\n")
+                            "'static' directory? [y|n]\n")
             if ans == 'y':
                 force = True
 
         if force:
             distutils.dir_util.remove_tree(os.path.join(os.getcwd(),
-                                                            'static'))
+                                                        'static'))
             print "All Done!"
         else:
             print "aborted!"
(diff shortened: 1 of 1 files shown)


sahilshekhawat:
There are some UI issues but I will fix them in #183

moorepants:
For this we need to see if we can support both IPython 2 and 3. It should be possible.

We don't necessarily have to support the first IPyhon 2.0.0 release but as long as we support at least the current version of 2.3.X that is probably fine.

sahilshekhawat:
Yeah it is possible and fairly simple.

moorepants:
Ok, please try to get this to work on IPython 2 and 3. Thanks.

pydy/viz/scene.py
```diff
@@ -472,17 +472,17 @@ def display_ipython(self):
...
-                                                              description=str(var))
+            self._widget_dict[str(var)] = widgets.FloatText(value=init_val,
+                                                        description=str(var))
             components.append(self._widget_dict[str(var)])
 
-        self.button = widgets.ButtonWidget(description="Rerun Simulations")
+        self.button = widgets.Button(description="Rerun Simulations")
         def button_click(clicked):
-            self.button.add_class('disabled')
+            self.button._dom_classes = ['disabled']
```
moorepants: Is it required to use these private methods? If we can avoid using them that would be ideal because they are bound to change in IPython 4.0.

sahilshekhawat: They suggested this way to do something they don't support http://ipython.org/ipython-doc/3/whatsnew/version3_widget_migration.html
and http://comments.gmane.org/gmane.comp.python.ipython.devel/14879
I don't think there is any reason we shouldn't. Also, we are including our custom html. Thus, thinkg which ipython support are not useful. They provide them to replace css and other properties.

moorepants: Ok, that's unfortunate. The common idiom in Python is to underscore private methods and in general it is not a good idea to rely on those methods. But that is what they recommend, so there may not be much choice.

sahilshekhawat:
It is now working with IPython >= 2 but I am not sure if this is the correct way.

pydy/viz/scene.py
```diff
@@ -24,6 +24,12 @@
     from IPython.html import widgets
     from IPython.display import display, Javascript
 
+    # Support for ipython 2 and ipython 3
+    if IPython.__version__[0] == '2':
```
moorepants: Here we may want to do something like:

```python
if ipy_ver < '2.0':
    raise Error
elif ipy_ver >= '2.0' and ipy_ver < '3.0':
    ipython_less_than_3.0 = True
else:  # ipython >= 3.0
   ipython_less_than_3.0 = False

Comment (moorepants) [inline pydy/viz/scene.py]: Also, since we require setuptools, make use of `pkg_resources.parse_version`.

File: pydy/viz/scene.py [hunk 3]
```diff
@@ -500,24 +519,37 @@ def button_click(clicked):
...
 
             self.button.description = 'Rerun Simulation'
 
-
         self.button.on_click(button_click)
         #components.append(button)
         html_file = open("static/index_ipython.html")
-        self.html_widget = widgets.HTMLWidget(value=html_file.read().format(load_url='static/' + self.scene_json_file))
+        if ipython2:
+            self.html_widget = widgets.HTMLWidget(value=html_file.read().format(load_url='static/' + self.scene_json_file))
```
moorepants: This line needs to be wrapped.

pydy/viz/scene.py
moorepants: There is one more issue here. The code currently supports not having IPython installed, i.e. IPython is an optional dependency. Either the code needs to take that logic into account or we need to make IPython a hard dependency.

If we go with the former, `display_ipython()` needs to raise an error if IPython is not installed.

pydy/viz/scene.py
```diff
@@ -472,17 +478,30 @@ def display_ipython(self):
...
+            else:
+                self._widget_dict[str(var)] = widgets.FloatText(value=init_val,
+                                                        description=str(var))
             components.append(self._widget_dict[str(var)])
 
-        self.button = widgets.ButtonWidget(description="Rerun Simulations")
+        if ipython2:
+            self.button = widgets.ButtonWidget(description="Rerun Simulations")
+        else:
+            self.button = widgets.Button(description="Rerun Simulations")
```
moorepants: Two blank lines before functions.

pydy/viz/scene.py
```diff
@@ -500,24 +519,37 @@ def button_click(clicked):
...
+            if ipython2:
+                self.button.remove_class('disabled')
+            else:
+                self.button._dom_classes = ['enabled']
 
             self.button.description = 'Rerun Simulation'
 
-
         self.button.on_click(button_click)
         #components.append(button)
```
moorepants: Remove this line.

pydy/viz/scene.py
moorepants: Make sure to run a linter on your code locally. I use pyflakes and pep8.

moorepants:
This looks like the correct way to handle things.

moorepants:
I tried this on IPython 3 and it works. The rerun simulation seems to work too. One issue that may need to be addressed in a new PR is that the constants are mixed up on load:
selection_087
g is set to 9.81 in the code and it gets mixed up with the other two in the JS side.

Also, this examples needs to make the sphere and cylinder radius's a function of the pendulum length so that when you change l the picture updates correctly.

sahilshekhawat:
Oh! thanks jason for pointing this things out. I am fixing them.

moorepants:
I was thinking about

pydy/viz/scene.py
sahilshekhawat: I don't think we need to make ``IPython`` a hard dependency. We can just raise a warning that IPython related functinalities will not be available.

moorepants: Ok, sounds fine to me. I agree.

sahilshekhawat:
Do you mind if I fix that mixed constants issue along with UI? I will be easier for me to fix all UI related bugs in a single PR

moorepants:
That's fine.

pydy/viz/scene.py
```diff
@@ -412,9 +426,10 @@ def create_static_html(self, overwrite=False, silent=False):
         simulation_outfile.write(json.dumps(self._simulation_data_dict, indent=4,
                                 separators=(',', ': ')))
         simulation_outfile.close()
-        if not silent: print("To view the visualization, open {}".format(
-                            os.path.join(dst, 'index.html')) +
-                            " in a WebGL compliant browser.")
+        if not silent:
+            print("To view the visualization, open {}".format(
```
moorepants: This no longer works. The user must view the html file using a server, right? If so we need to have this message tell how to do that.

pydy/viz/scene.py
```diff
@@ -433,13 +448,13 @@ def remove_static_html(self, force=False):
 
         if not force:
             ans = raw_input("Are you sure you would like to delete the " +
-                                "'static' directory? [y|n]\n")
+                            "'static' directory? [y|n]\n")
```
moorepants: The directory is called `static` in the source tree and that's fine. But maybe it should be called something like `pydy_app` or something when it is copied to cwd. I feel like a more unique name is in order.

pydy/viz/scene.py
```diff
@@ -465,59 +480,105 @@ def display(self):
...
         versions>=2.0.0
 
         """
 
+        # Raise error whenever display_ipython() is called
+        # and IPython is not installed or IPython < '2.0.0'
+        if IPython is None:
+            raise ImportError('IPython is not installed but is required. ' +
+                              'Please installed IPython >= 2.0 and try again')
+        elif ipython_less_than_3 is None:
```
moorepants: You could probably just have one variable here. If Python < 2.0 is installed or IPython isn't installed then we just make `has_ipython=False`. Is it necessary to have two variables to check this?

moorepants: But maybe the two different error messages are nice for the user. This probably doesn't matter either way.

pydy/viz/scene.py
sahilshekhawat: I have addressed ``create_static_html()`` in my another PR. I was fixing the formatting but did it a bit over.

pydy/viz/scene.py
sahilshekhawat: So,  should I remove it or is it fine?

pydy/viz/scene.py
moorepants: Ok, You can change this in the other PR.

pydy/viz/scene.py
moorepants: I think you may need both variables because of the checks below. It's fine.

sahilshekhawat:
@pydy/pydy-developers This PR is ready for final review.

pydy/viz/scene.py
```diff
@@ -465,59 +480,105 @@ def display(self):
...
+        IPython notebook. It is only supported by IPython
         versions>=2.0.0
 
         """
 
+        # Raise error whenever display_ipython() is called
+        # and IPython is not installed or IPython < '2.0.0'
+        if IPython is None:
+            raise ImportError('IPython is not installed but is required. ' +
+                              'Please installed IPython >= 2.0 and try again')
```
moorepants: "Please installed" > "Please install"

moorepants:
I tried out the double pendulum notebook example with IPython 2.4.1 and:
- The animation slider does not seem to work. It doesn't move when play is pressed and can't be moved manually.
- It does not seem like the "rerun simulation" button is working.

moorepants:
I think this is good to go in. We can address the other IPython 2 anomalies in new PRs. It has basic functionality in IPython 2.

PR #196, opened 2015-05-29
Fixed static directory creation issue on scene.display()

### LINKED ISSUE #156: Scene.create_static_html() may be useless now.
Previously you could export all of the html/css/js files to a directory and simply open up the index.html file in a web browser to interact with the visualization. Now that doesn't seem to be the case and a server is required due to AJAX?

We need to either update the docs explaining how to use the static files or make this work again.

### LINKED ISSUE #188: The static directory should not be created on every call to `Scene.display()`.
Every time `Scene.display()` is called it creates a `static` directory and two json files. If the `static` directory is already present in the directory the user is asked wheter to delete the contents or not.

Why do we need to copy the static html/css/js files to every directory? Shouldn't these be able to be loaded from where they are installed to? I thought the only reason to copy the static files down where to be able to make the visualization relocatable (i.e. to post on a website or something). I don't think these need to be copied into the current directory unless the user explicitly asks to do that. See issue #156 for more.

sahilshekhawat:
Fixes #156
Fixes #188

  commit message.
  in the unit test.)
  follow deprecation cycles.)


sahilshekhawat:
shouldn't we delete extra json files before starting the visualisation? We already have download json functionality. User can download the json if he wants to.

moorepants:
Can you explain what you've implemented in the PR?

sahilshekhawat:
Now, by default static files from pydy are used, they are not copied to the current directory. replaced overwrite parameter with copy which tells whether to copy to current directory or not and some docstring changes

pydy/viz/scene.py
```diff
@@ -363,14 +363,10 @@ def generate_visualization_json_system(self, system, **kwargs):
...
                                           system.constants.values(), **kwargs)
 
-    def create_static_html(self, overwrite=False, silent=False):
-
-        """Creates a directory named ``static`` in the current working
-        directory which contains all of the HTML, CSS, and Javascript files
-        required to run the visualization. Simply open ``static/index.html``
-        in a WebGL compliant browser to view and interact with the
-        visualization.
+    def create_static_html(self, copy=False, silent=False):
```
moorepants: In general we need to avoid changing call signatures of public functions. Any time this is done we have to deprecate the behavior.

pydy/viz/scene.py
```diff
@@ -363,14 +363,10 @@ def generate_visualization_json_system(self, system, **kwargs):
...
 
+        """
+        Copies json files into a static directory, pydy/viz/static by default,
```
moorepants: I thought this copied files into the current working directory.

moorepants:
@sahilshekhawat I'm not sure this is what we want to do. The method `create_static_html()` should do what it says. It should create a `static` directory in the current working directory. This static directory should be a replica of the static directory in source tree.

The `display()` method should not be calling `create_static_html()` if should simply change into the appropriate directory and run the server.

So essentially much of the logic you have in `create_static_html()` needs to be moved into `display()`.

Also the `overwrite` kwarg should not be changed and we should leave the public API behavior and call signature in tact.

Lastly make sure that you have unit tests for any new functionality.

pydy/viz/scene.py
sahilshekhawat: It does but only if ``copy=True`` is passed but by default it only copies into ``pydy/viz/static``

sahilshekhawat:
Oh! I get it now, So, we want create_static_html() to create a static directory but scene.display() should not call it and thus, should not create the directory.
Okay, I am on it. Thanks for clearing that for me.

pydy/viz/scene.py
moorepants: This needs to stay as `overwrite` not `copy`.

pydy/viz/scene.py
moorepants: What do you mean "copies into `pydy/viz/static`"? Isn't it supposed to copy the directory `/where/pydy/is/installed/pydy/viz/static` to the current working directory?

pydy/viz/scene.py
```diff
@@ -378,43 +374,42 @@ def create_static_html(self, overwrite=False, silent=False):
 
         Parameters
         ----------
-        overwrite : boolean, optional, default=False
+        copy : boolean, optional, default=False
```
moorepants: overwrite needs to be retained and I don't think copy is needed at all.

pydy/viz/server.py
```diff
@@ -5,12 +5,16 @@
 from SimpleHTTPServer import SimpleHTTPRequestHandler
 
 
-def run_server(port=8000,scene_file="Null"):
-    #change dir to static first.
-    os.chdir("static/")
+def run_server(port=8000, scene_file="Null"):
+    # change dir to static first.
+    if os.path.exists(os.path.join('static', scene_file)):
```
moorepants: I don't think we need this either. The server should always run the viz from the files in the source tree.

sahilshekhawat:
Forgot to mention, this PR is a work in progress. I totally misunderstood what was needed to fix. I will update it with a correct fix.

moorepants:
@sahilshekhawat Do you need any help with this?

sahilshekhawat:
No, I was just working on the API. Will take some time out today to complete it.

moorepants:
@sahilshekhawat This has merge conflicts with master now.

sahilshekhawat:
Can I just close this one and open a new one? I want to revert those commits anyway.

moorepants:
You can, but I don't think it is necessary (to open a new one or revert commits).

moorepants:
@sahilshekhawat Please review #201 and merge before this one.

moorepants:
@sahilshekhawat What is your plan here? Let's try to get this merged in next.

sahilshekhawat:
Yeah, I have started working on it.

sahilshekhawat:
Visualization json files are created twice, one at the current directory and other in the static directory, is there any reason we do that? and also, should we automatically delete json files, since user always have the option to download the json.

moorepants:
I don't know why they are created twice. That doesn't sound right.

Here are few problems I've noticed about these files (not sure of solution yet):
1. The fill up the cwd and you have no idea what the differences are. Many times they are just duplicates of each other.
2. There is no reason to have 2 files with the current way they are designed. The two files are linked by the unique object IDs, so you can't have multiple simulation files for one scene file, or vice versa. I thought the reason there were two files was to support that.
3. Even if you wanted to load different scene/sims from the GUI the filenames do not give you enough useful information to choose the one you may want.
4. If you are in the notebook and rerun simulations from the GUI, where do these files go? The file list in the "load sim" box doesn't seem to change with each rerun.

One solution, would be to hide the files from the user and remove the "load simulation" box/button from the GUI. Then you have to call `display()` or press "rerun" every time you want a new simulation. The json files should someone be cleaned up when they are no longer in use. Maybe they could be deleted as soon as the GUI loads them. We can then come back at a later date and properly design this.

chrisdembia:
The files could be written to `/tmp`.

moorepants:
Yes, `/tmp` would be fine with me.

moorepants:
If you use `/tmp` make sure to use the tempfile module in Python so that it is cross platform.

sahilshekhawat:
This is ready for review. Thanks

oliverlee:
I recommend that you rebase your commits and squash/fixup your revert/merge cleanup commits. It provides

More information here: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request 
and here: https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase-i

sahilshekhawat:
I have merged master twice into this branch. rebasing results in many merge conflicts. Can I skip it this time? Will take care of this from now on.

moorepants:
Yes, we aren't too stick on having a super clean history. I actually prefer the merging technique to keep up with master because it doesn't change the public history and merge conflicts are way easier to deal with.

oliverlee:
I also think that changing public history is a terrible no good very bad idea but I also don't consider topic branches to be public.

moorepants:
Yes, they are not public in the sense that they are not yet in the master branch. But for larger projects when multiple people may be cloning your PR and maybe adding commits to it, your rebasing will screw things up for them. So in that sense it is public as soon as you share for collaboration.

But more importantly is that rebasing with merge conflicts is a nightmare process and merging isn't.

Regardless, if you think that rebasing is necessary for this PR then you can ask Sahil to do it, but our current dev process defaults to the SymPy workflow which recommends merging over rebasing.

moorepants:
This looks fine but there are still calls to `create_static_html` in `display_ipython`, right?

And we should remove the GUI element for loading simulation files here. That way after this PR is merged the simulation and scene files will be completely hidden from the user. If you have time to enable the loading/saving, etc of these files before the release then you can add it back in a new PR.

sahilshekhawat:
Yes, display ipython calls ``create_static_html``.

sahilshekhawat:
Should I remove loading simulation file from ipython too?

moorepants:
I think we should completely remove loading the sim files from this PR.

And actually the GUI button for viewing and downloading the scene file is sorta useless too.

sahilshekhawat:
This is ready for a final review. I have removed loading simulation file option.

sahilshekhawat:
@pydy/pydy-developers This is ready for review. Thanks

moorepants:
I'd really like to get this in but it doesn't seem like you fixed this for the creation of the static directory when calling `display_ipython`, I am wrong?

sahilshekhawat:
Fixing.

sahilshekhawat:
the issue is that when we include our html file in ipython notebook it references static files as ``static/*`` thus, static folder must be in the same directory as the notebook file.

moorepants:
If it absolutely has to be in the same directory at least make it a hidden directory and prevent the json files from filling up the cwd.

sahilshekhawat:
okay

pydy/viz/scene.py
```diff
@@ -459,8 +461,15 @@ def remove_static_html(self, force=False):
...
         """Displays the scene in the default webbrowser."""
-        self.create_static_html()
-        run_server(scene_file=self._scene_json_file)
+        tmp_dir = tempfile.mkdtemp(prefix='pydy_', suffix='_viz')
+        src = os.path.join(os.path.dirname(__file__), 'static')
+        distutils.dir_util.copy_tree(src, tmp_dir)
+        shutil.move(os.path.join(os.getcwd(), self._scene_json_file),
+                    os.path.join(tmp_dir, self._scene_json_file))
+        shutil.move(os.path.join(os.getcwd(), self._simulation_json_file),
+                    os.path.join(tmp_dir, self._simulation_json_file))
```
moorepants: I'm not sure about this method. The `generate...json()` method dumps the files into the cwd. That method should be dumping them into this tmp directory.

With this current method, if you call `generate_visualization_json()` multiple times you will fill the cwd with json files and then if you call `display()` you'll only move the last json file to the tmp dir. So this doesn't really solve the problem.

pydy/viz/server.py
```diff
@@ -1,29 +1,77 @@
...
+    ----------
+    port : integer
+        Defines the port on which the server will run.
+    scene_file : name of the scene_file generated for visualization
+        Used here to display the url
+    directory : path of the directory which contains static and scene files.
+        Server is started in this directory itself.
+
+    """
+    def __init__(self, port=8000, scene_file="Null", directory="static/"):
```
moorepants: Shouldn't the scene file and directory be required args, not kwargs? And `"Null"` is not typical python practice. You should use `None` in cases like that.

`def __init__(scene_file, directory, port=8000):`

`scene_file` should be the "path to the scene file".

moorepants: Or is it only relative to the directory?

moorepants:
So after thinking about this some, maybe we need to rethink the `Scene` class. It really isn't designed so well. But it was also released in some form in PyDy 0.2...

A scene has visualization frames, cameras, lights, and a reference frame/point.

`scene = Scene(frame, point, *viz_frames, cameras, lights)`

You then can add in the the trajectories of the visualization frames, cameras, and lights by passing in the numerical values of the coordinates and constants as a function of a time array:

```python
scene.time = time_array  # ndarray, shape(n,)
scene.coordinate_trajectories = coord_traj  # ndarray, shape(n, m) or dict mapping SymPy functions to 1d arrays.
scene.constants_values = cons_map  # dict mapping SymPy symbols to floats 
 time, coordinates, constants)
```

Or you can provide a `System` which generates the time, coordinate, and constants values:

`scene.system = system`

But you shouldn't be able to provide both. If a System is provided you should be prevented from adding: time, coordinate_trajectories, or constants_values. And if you provide any of the three individual things you shouldn't be able to provide a system.

Finally, you can display the scene with:

```python
scene.display()
```
or
```python
scene.display_ipython()
```

All of the "generate*" methods should be private so that we can control completely the creation of the json files and keep them hidden from the user.

moorepants:
Sahil, chat me on gitter when you get this.

sahilshekhawat:
@oliverlee This PR might be closed but I want server stuff to be merged. Should I raise another PR addressing your comments?

moorepants:
@sahilshekhawat Just fix those comments here. I'll look over this in a few minutes.

oliverlee:
@sahilshekhawat defer to @moorepants

moorepants:
@sahilshekhawat Earlier on chat you told me that you had made more changes to this branch but I don't see them. For example you said you made the generate methods private.

How much more work do you have on Scene? If not much more I may just open a new PR. If that happens you should open a new PR for the server code separately.

sahilshekhawat:
There are some bugs left. Should I just push it as it is so that you can see how much I have done?

moorepants:
Yes, please. And maybe it is the right choice to move the server changes in to a new PR.

sahilshekhawat:
I have not made generating json private yet but I think somthing like:
```
scene.display(sys)
```
and display_ipython will have a call to generate json files.

moorepants:
I was thinking of allowing the user to pass in the system to the `__init__` method or set it as an attribute. We also have to support passing in the time, constants, and trajectories which can also be passed into `__init__`. But I could see how this makes sense: `scene.display(trajectory)` or `scene.display(simulation_results)`.

moorepants:
I'm going to push my scene code into a new pr so you can see what I have in mind.

sahilshekhawat:
I was just going to ask you, Thanks

pydy/viz/server.py
sahilshekhawat: It runs from static directory present in current directory but Yeah, i too think that this should be a required arg.
Also, scene_file is used to calculate the url and print it, as a result it can't be None, It must be Null (a string)

sahilshekhawat:
Closing in favor of #242

PR #210, opened 2015-06-09
Ensures that constants are ordered correctly in GUI.

moorepants:
Fixes #204

I've tested this on the mass_spring_damper example and it works correctly now. There is also a ton of code clean up and new tests for Scene. I also removed the width and height attributes from Scene as they were not used anywhere.

- [x] There are no merge conflicts.
- [x] If there is a related issue, a reference to that issue is in the
  commit message.
- [x] Unit tests have been added for the bug. (Please reference the issue #
  in the unit test.)
- [x] The tests pass both locally (run `nosetests`) and on Travis CI.
- [x] The code follows PEP8 guidelines. (use a linter, e.g.
  [pylint](http://www.pylint.org), to check your code)
- [x] The bug fix is documented in the [Release
  Notes](https://github.com/pydy/pydy#release-notes).
- [x] The code is backwards compatible. (All public methods/classes must
  follow deprecation cycles.)
- [x] All reviewer comments have been addressed.

Issue #207, opened 2015-06-08
Custom camera orientation does not seem to work


moorepants:
If you run the mass_spring_damper example there is a custom camera that used to rotate the scene properly. See the first screenshot in this PR #97. But for some reason only the location of the camera seems correctly adjusted and not the orientation.

sahilshekhawat:
This issue must be fixed in #183.

moorepants:
It does not seemed to be fixed in #183 for me. The orientation of the camera in the mass_spring_damper should be such that the mass bobs up and down.

moorepants:
@sahilshekhawat Would you mind looking into this issue. I don't think it is on the Python side. I haven't been able to find the bug yet in the JS.

sahilshekhawat:
I am currently working on fixing issues with camera.

sahilshekhawat:
Okay, So I spent about one and a half day figuring out the reason and i still don't know.
``camera.applyMatrix()`` must work as it sets the value of ``position``, ``quaternions`` and ``scale`` but it doesn't changes quaternions only. Position is changed. Moreover, When I manually set those values e.g.
```
var position = new THREE.Vector3();
var quaternion = new THREE.Quaternion();
var scale = new THREE.Vector3(1, 1, 1);
initMatrix.decompose(position, quaternion, scale);
_camera.position = position;
_camera.quaternion = quaternion;
_camera.scale = scale;
```
then again quaternion's value is not changed.
Correct orientation works when we do
```
_camera.matrixAutoUpdate = false;
```
which doesn't changes the value of transformation matrix from incorrect quaternions but then again trackballcontrols are not working due to that.
Whenever we run ``camera.updateMatrix()`` it changes the orientation which makes it sure the bug is in setting the value of ``camera.quaternions``
I am leaving it until the end and moving on to fix other bugs. Will look into it later.

moorepants:
@tarzzz Do you have any idea why this is not working? It used to work just fine (at least at the end of last summer).

Sahil is going to file an issue on threejs about this.

sahilshekhawat:
Instead of an issue, I have asked on stackoverflow: http://stackoverflow.com/questions/30818179/rotation-does-not-change-after-setting-transformation-matrix-for-camera

tarzzz:
Did a little brainstorming. When the frame is rotated 90degrees with `ceiling.y` as the Axis, it behaves as per the expectation.

[...]

PR #211, opened 2015-06-10
Adds a test for the OrthoGraphicCamera dict generation.

moorepants:
This was a test for issue #207, but it ended up uncovering a bug in the
OrthoGraphicCamera. PR #210 needs to be merged before this one.

  commit message.
  in the unit test.)
  follow deprecation cycles.)

### DIFF
--- a/pydy/viz/tests/test_scene.py
+++ b/pydy/viz/tests/test_scene.py
@@ -4,7 +4,7 @@ import os
 import shutil
 import glob
 
-from numpy import linspace
+import numpy as np
 from numpy.testing import assert_allclose
 from sympy import symbols
 import sympy.physics.mechanics as me
@@ -12,7 +12,7 @@ import sympy.physics.mechanics as me
 from ...system import System
 from ..shapes import Sphere
 from ..visualization_frame import VisualizationFrame
-from ..camera import PerspectiveCamera
+from ..camera import PerspectiveCamera, OrthoGraphicCamera
 from ..light import PointLight
 from ..scene import Scene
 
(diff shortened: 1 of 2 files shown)


moorepants:
@pydy/pydy-developers Ok, this is ready for review now that #210 is in.

pydy/viz/tests/test_scene.py
```diff
@@ -207,6 +207,39 @@ def test_generate_scene_dict(self):
...
+
+        expected_orientation_matrix = np.array([0.0, 0.0, 1.0, 0.0,
+                                                0.0, -1.0, 0.0, 0.0,
+                                                1.0, 0.0, 0.0, 0.0,
+                                                30.0, 0.0, 0.0, 1.0])
+
+        assert_allclose(scene_dict['cameras'][camera_id]['init_orientation'],
+                        expected_orientation_matrix, atol=1e-14)
+
+        assert scene_dict['cameras'][camera_id]['type'] == 'OrthoGraphicCamera'
```
sahilshekhawat: I am just wondering if we can add some more ``asserts`` here or just checking the type is enough? Since we are using ``id(self)`` it can be same for two classes during different life times.

moorepants: Right now there are no tests for camera.py, see #212. I just created this one to try to find out where #207 was going wrong. 

Plus the full scene dict is checked already in the other test. So some here will be redundant.

We will need to add full tests for camera.py. I just wrote this one and though we could merge it.

sahilshekhawat: But I am not sure if someone can actually define two cameras. I think this should be enough.

sahilshekhawat: Okay, Thanks

sahilshekhawat:
works, :+1: Lets merge this one.

moorepants:
Remember to check of the boxes at the top as a reviewer on the future PRs.

pydy/viz/tests/test_scene.py
moorepants: Yes, I do not know why you can define two cameras? I'm guessing the intention was that you can define multiple cameras for a scene and then switch among them from the GUI.

Issue #136, opened 2015-02-09
Support output equations in the code generators.


moorepants:
Support output equations: both those that share evaluation at each time step and those that can be evaluated after the simulation.

Many times you need to compute some function of the states, parameters, and time. The user may desire to do this after the simulation or during it. In general lambdify is probably fine if you just need to evaluate some expressions after the integration but it may not be suitable during the integration due to speed reasons. This would be very useful for controllers and for visualization purposes. Additionally many of the expressions may share common sub expressions to those in the EoMs.

The `eval_arrays` function should support an output matrix.

```python
g = CythonODEFucntionGenerator(rhs, coordinates, speeds, constants, outputs=symbolic, output_matrix)
rhs = g.generate()
xdot, y = rhs(x, t, p)
```
It would be neat to support having the output, y, as an input to the user defined specifieds function:
```python
def eval_specifieds(x, t, p, y):
    return some_function_of_the_outputs
```
That way the user could write a controller for example that uses the outputs in a numerical function instead of being forced to write it symbolically apriori.

It is also useful to evaluate a the outputs after integration in a fast loop over all the states so it could be worth decoupling the output eval from the state derivative eval too, even though they may share common sub expressions.

moorepants:
y(t, x, r, p): nonlinear in only the state
kinetic energy $y = \frac{1}{2}mv^2$
holonomic constraint residual: $y = l*cos(q1) - d*sin(q3)$
solved acceleration $y = a = (-cv - kx + f)/m$
feedback controller: $y = -k(x - x_c)$

y(t, x, x', r, p): nonlinear in the state and its time derivative
example: sum of squared acceleration measure numbers $y = \dot{u}_1^2 + \dot{u}_2^2$

y(t, x, x', r, p): linear in x'
constraint force: $a\dot{u}_1 - T_1 + b\dot{u}_2 - T2 = F$
constraint force from lagrange multipliers
udots: y = \dot{u} = M^{-1} F


Here is a fully symbolic example showing a set of things you may like to calculate for a time instance:

```python
import sympy as sm
import sympy.physics.mechanics as me

m1, m2, l1, l2, c, g = sm.symbols('m1, m2, l1, l2, c, g')
q1, q2, u1, u2, T1, T2 = me.dynamicsymbols('q1, q2, u1, u2, T1, T2')
u3, u4 = me.dynamicsymbols('u3, u4')

N = me.ReferenceFrame('N')
A = me.ReferenceFrame('A')
[...]

Issue #44, opened 2014-04-14
Example Improvement with gallery


moorepants:
- Include examples in an example "gallery" in the docs. Ideally each example has a visualization of the motion.
- Examples should be run as unit tests, or at least doc tests. It would be nice to have the code for the example as separate and executable but also embedded in the docs (maybe even tutorial style).

chrisdembia:
@jcrist with @moorepants

jcrist:
Based on the folders in `/examples`, here's what we need to do:

- [ ] babyboot
- [ ] body_with_arm
- [ ] chaos_pendulum
- [ ] differential_drive
- [ ] double_pendulum
- [ ] inertia_of_torus
- [ ] Kane1985
- [ ] language_templates
- [ ] mass_spring
- [ ] mass_spring_damper
- [ ] npendulum
- [ ] parallel-execution
- [ ] rattleback
- [ ] rollingdisc
- [ ] sliding_ladder
- [ ] three_link_conical_pendulum
- [ ] tutorials

jcrist:
Edit the above comment to dibs one, then x it when it's done.

jcrist:
## Proposed format

Format:

1. Title
2. Problem Setup

    This should include a vector drawing of the system (see the differential drive or n-pendulum examples, as well as a basic description of said drawing, and what we're trying to solve for.

3. Build System

    Whatever title headers you want here. Should lead through the development and solution of the EOM. Be descriptive, but not verbose. Slightly better than well commented code.

4. Simulation

    Optional. If used, should include pretty matplotlib plots, or a simulation with the viewer (never used this, not sure what to do with that).

### Style:

Because `nbconvert` does some fun things with html tags, a uniform style might be nice. 

- Continuous sections of prose should be merged into the same cell (don't have two adjacent text cells). 
- Title should be heading 1 (single `#`) and in its own cell. 
- Subheaders should be included in the same cell as any following text, and should progress logically down the heading hierarchy (subsection under heading `n` with heading `n + 1`).

chrisdembia:
+1

moorepants:
+1

Talk to Tarun, but the ideal situation is that if the notebook has a visualization, nbconvert would somehow retain it (the visualizations are not necessarily tied to the "server" side) so nbviewer, for example, should be able to show it.

moorepants:
[...]

PR #217, opened 2015-06-10
Fix bug 215

### LINKED ISSUE #215: rattleback ellipsoid_noslip example raises exception
```python
23:42 ~/r/p/e/rattleback (master *=) ipython ellipsoid_no_slip.py 
TypeError                                 Traceback (most recent call last)
/Users/chris/repos/pydy/examples/rattleback/ellipsoid_no_slip.py in <module>()
     73 
     74 # Forces and Torques
---> 75 F_P = sum([cf*uv for cf, uv in zip(CF, Y)])
     76 F_RO = m*g*Y.z
     77 T_R = -s*R.ang_vel_in(N)

/Library/Python/2.7/site-packages/sympy/physics/vector/vector.pyc in __add__(self, other)
     63     def __add__(self, other):
     64         """The add operator for Vector. """
---> 65         other = _check_vector(other)
     66         return Vector(self.args + other.args)
     67 

/Library/Python/2.7/site-packages/sympy/physics/vector/vector.pyc in _check_vector(other)
    669 def _check_vector(other):
    670     if not isinstance(other, Vector):
--> 671         raise TypeError('A Vector must be supplied')
    672     return other

TypeError: A Vector must be supplied
```

oliverlee:
Fixes #215.

All python files run and cpp executables build. I did not test the executables themselves.

Please review the changes to the python files in commit a2d5cb2947eab8b61ad894fbe5dcef7677ef9c9a

The range of the double for loops was reduced from 8 to 5 as @hazelnusse probably intended to do so in  this change 940031574e9ecfd8cce32257892f13eff809742d but forgot. The commit shows that the size of the state is reduced from 8 to 5.

I also set the velocity of points O and RO so please verify that the values are correct.

In the second commit 64afea271431235fcd9109dfe5565b98d844b20d
It looks like @hazelnusse forgot to add some source files so I commented out the relevant section in CMakeLists files.

  commit message.
  in the unit test.)
  follow deprecation cycles.)

### DIFF
--- a/examples/rattleback/cpp/CMakeLists.txt
+++ b/examples/rattleback/cpp/CMakeLists.txt
@@ -41,8 +41,9 @@ target_link_libraries(paraboloid_sim
                       ${GSL_LIBRARIES})
 
 add_executable(ellipsoid_equilibrium ellipsoid_eq.cc)
-target_link_libraries(ellipsoid_equilibrium rattleback vtkRendering)
+target_link_libraries(ellipsoid_equilibrium rattleback ${VTK_LIBRARIES})
 
-add_executable(RattleBackAnimation RattleBackAnimation.cc
-                                   RattleBackCallback.cc)
-target_link_libraries(RattleBackAnimation rattleback vtkRendering)
+# rattleback animation source files are missing:
+#add_executable(RattleBackAnimation RattleBackAnimation.cc
+#                                   RattleBackCallback.cc)
+#target_link_libraries(RattleBackAnimation rattleback ${VTK_LIBRARIES})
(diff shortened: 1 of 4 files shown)


oliverlee:
Does Travis do anything? It might be nice to have it run examples in the future.

moorepants:
Yes, that is on the todo list. See: https://github.com/pydy/pydy/issues/44

moorepants:
This looks fine to me, just to get the Python code running again. We can merge as is, but just for the record these things need to be addressed in the future once we get all the example running.

- The C++ files need to be updated with the results of the python files. Ideally we don't commit any of the generated C++ files.
- The PyDy code generators should be able to handle this example. We don't have C++ code generation yet and we don't support output equations yet, see #136.
- A figure is needed.
- There should be a top level `run.py` (or make/cmake file) that runs this entire example from the raw code down to the plot generation.
- The tex file seems like the start of an academic paper. This should be turned into some rst that could be displayed in the example gallery.

I'm going to merge as is and we can try to get the examples overhauled at the SciPy 2015 sprints.

moorepants:
We can document in the release notes that examples were updated at a later PR.

Issue #92, opened 2014-08-16
Don't show rerun option if visualization wasn't built with a System


moorepants:
If you manually call generate_viz_json without using a System class in the notebook the rerun button shows up, but that won't work without a System class.

moorepants:
`display_ipython()` may work properly with or without a System. This needs to be investigated.

moorepants:
I have confirmed in the human standing tutorial that "Rerun simulations" does not seem to work if you do not have a System.

PR #224, opened 2015-06-11
Rerun simulation option disabled if no System present.

moorepants:
Fixes issue #92.

- Reorganization and cleanup of display_ipython().
- Rerun button classes are changed properly on click.

# Tasks
  commit message.
  in the unit test.)
  follow deprecation cycles.)

### DIFF
--- a/pydy/viz/scene.py
+++ b/pydy/viz/scene.py
@@ -9,6 +9,7 @@ import distutils.dir_util
 import datetime
 from collections import OrderedDict
 from pkg_resources import parse_version
+
 # external
 from sympy.physics.mechanics import ReferenceFrame, Point
 
@@ -85,6 +86,7 @@ class Scene(object):
         self.reference_frame = reference_frame
         self.origin = origin
         self.visualization_frames = list(visualization_frames)
+        self._system = None
 
         try:
             self.name = kwargs['name']
@@ -94,20 +96,19 @@ class Scene(object):
         try:
             self.cameras = kwargs['cameras']
         except KeyError:
+            camera_point = self.origin.locatenew('p_camera', 10 *
+                                                 self.reference_frame.z)
             self.cameras = [PerspectiveCamera('DefaultCamera',
-                            self.reference_frame,
-                            self.origin.locatenew(
-                                'p_camera',
-                                10*self._reference_frame.z))]
+                                              self.reference_frame,
+                                              camera_point)]
 
         try:
             self.lights = kwargs['lights']
         except KeyError:
-            self.lights = [PointLight('DefaultLight',
-                           self.reference_frame,
-                           self.origin.locatenew(
-                               'p_light',
-                               10*self._reference_frame.z))]
+            light_point = self.origin.locatenew('p_light', 10 *
+                                                self.reference_frame.z)
+            self.lights = [PointLight('DefaultLight', self.reference_frame,
+                                      light_point)]
 
     @property
     def name(self):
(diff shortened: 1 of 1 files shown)


moorepants:
@sahilshekhawat Can you review this PR?

moorepants:
@pydy/pydy-developers This PR is ready for review.

sahilshekhawat:
Just curious, is it okay to define instance attributes outside __init__?

pydy/viz/scene.py
```diff
@@ -459,33 +458,59 @@ def remove_static_html(self, force=False):
...
+            self._rerun_button._dom_classes = ['btn-info', 'active', 'disabled']
+
+        self._rerun_button.description = 'Rerunning Simulation...'
+
+        original_scene_file = self._scene_json_file
+        try:
+            self._system.constants = {s: w.value for s, w in
+                                      self._constants_text_widgets.items()}
+            self.generate_visualization_json_system(self._system)
+        except:
```
sahilshekhawat: Can you please add error here.

moorepants:
You can define them any time you want. But if you are asking if it is good practice, well that depends on what you are doing. Here, we have a mutable object where (I think) all of the attributes created in the methods are "private" (in the Python sense). So users shouldn't even interact with them.

I don't think there is any reason not to and it is common among all of the Python code I've written.

pydy/viz/scene.py
moorepants: It is impossible to predict what the error would be. There are too many possibilities. I just want it to prevent hanging the callback if there is any error at all.

sahilshekhawat:
Oh! Okay.

pydy/viz/scene.py
moorepants: But maybe I need to reset the constants on system in the except clause.

moorepants: I have not updated this to fully revert things if the simulation with the new constants fail.

moorepants:
@sahilshekhawat Can you look over this once more and merge if you think it is acceptable?

moorepants:
@pydy/pydy-developers This PR needs to be reviewed. Thanks.

pydy/viz/scene.py
```diff
@@ -459,133 +458,147 @@ def remove_static_html(self, force=False):
...
         self.create_static_html()
-        run_server(scene_file=self.scene_json_file)
+        run_server(scene_file=self._scene_json_file)
 
-    def display_ipython(self):
-        """
-        Method to display the visualization inside the
-        IPython notebook. It is only supported by IPython
-        versions>=2.0.0
+    def _rerun_button_callback(self, clicked):
```
sahilshekhawat: clicked is unused.

sahilshekhawat:
It works, everything is fine. We can merge this one after removing ``clicked``.

pydy/viz/scene.py
sahilshekhawat: Is it necessary to include for ``on_click``?

moorepants: I think that is necessary for the callback to function. Let me check the IPython docs.

moorepants: This (https://ipython.org/ipython-doc/dev/api/generated/IPython.html.widgets.widget_button.html) says:

"The callback will be called with one argument, the clicked button widget instance."

sahilshekhawat: Yes, we need it. Thanks.

sahilshekhawat:
Everything is fine from my side. Can we merge this one?

moorepants:
Hold up. I'm fixing the button callback to be more correct.

moorepants:
Ok, you can check of any remaining boxes and merge.

Can we have a brief meeting on gitter?

moorepants:
Ok, this is in.

PR #226, opened 2015-06-12
Custom bootstrap to place nice with ipython notebook

### LINKED ISSUE #203: The notebook visualization may prevent you from restarting the kernel
I ran the mass spring damper example in PR #97 and if I "Run All" and then "Kernel>Restart" the notebook just hung in a grey screen. I had to close the browser tab and open a the notebook again to bring it back to life. There also seems to be odd behavior when "close and halt" is selected in a notebook with a viz present.

### LINKED ISSUE #209: Displaying the visualizer distorts the ipython header, etc.
calling display_ipython() causes the header to get unnecessary whitespace.

python 2.7.6 ipython 3.0.0

before
screenshot from 2015-06-08 23 25 06

after
screenshot from 2015-06-08 23 28 49

### LINKED ISSUE #227: The slider is very short
selection_092

sahilshekhawat:
bootstrap is already imported for ipython notebook, thus, we don't need to
Fixes #209 
Fixes #227 
Fixes #203

  commit message.
  in the unit test.)
  follow deprecation cycles.)


pydy/viz/static/index_ipython.html
```diff
@@ -10,7 +10,8 @@
     <link type="text/css" rel="stylesheet" href="static/css/main.css">
     <link type="text/css" rel="stylesheet" href="static/css/codemirror/blackboard.css">
     <link type="text/css" rel="stylesheet" href="static/css/slider.css">
-    <link type="text/css" rel="stylesheet" href="static/css/bootstrap.min.css">
+    <!-- commented import of bootstrap until css leak is fixed in Ipython widgets-->
+    <!--<link type="text/css" rel="stylesheet" href="static/css/bootstrap.min.css">-->
```
moorepants: Can we just remove it if it is no longer used?

moorepants: Why is it needed at all, even if the widgets support it?

sahilshekhawat: We don't know how or which version they are implementing. I

sahilshekhawat: I think to fix the leak issue they will have to fix it both ways.ie. widget's css should not be applied to the notebook's css  as well as notebook's css should not be applied to widget.

moorepants: The point I'm trying to make is "is there any reason not to delete this line?".

sahilshekhawat:
This needs to be review, thanks

moorepants:
This solves the notebook header issue but it seems like the bootstrap version distributed with IPython may not have everything that we have in our distribution. Note that the glyphs are all gone and the sizing of the canvas is even different.
selection_093
 
So, either we figure out how to readjust things to that we can use the IPython boostrap or we stick with ours. Maybe using less to adjust all the css could be worthwhile, but maybe not if it takes a long time.

Also, I can imagine that this could have issues in IPython 2, too as it likely has a different version of bootstrap.

sahilshekhawat:
Yes, trying less is probably a good idea.

sahilshekhawat:
I have compiled bootstrap from less and it works great. Also, I found out that our html fixes best when browser's zoom is at 100%, decreasing it leads to the screenshot you showed. I think its because everything is getting changed to pixels and decreasing or increasing zoom doesn't changes it but it remains responsive.

sahilshekhawat:
Also, After replacing our bootstrap, I am able to restart kernel with any issue.

moorepants:
Ok, so if you complied bootstrap and you then commit it to the repo as a custom bootstrap version, this fixes both issues?

That sounds fine.

We will eventually need to have the instructions in place for compiling, but I'd be fine if you want to commit these in.

pydy/viz/static/css/main.css
```diff
@@ -36,70 +36,74 @@ input[type=text]:focus{
...
 }
 
-ul {
+#container ul {
   list-style-type: none;
 }
+
+#container .form-control{
+	margin-bottom: 13px;
+}
```
moorepants: eof

pydy/viz/static/index.html
```diff
@@ -8,7 +8,7 @@
     <title>PyDy Visualizer</title>
 
     <!--CSS -->
-    <link type="text/css" rel="stylesheet" href="css/bootstrap.min.css">
+    <link type="text/css" rel="stylesheet" href="css/bootstrap.css">
```
moorepants: Is there a reason we can't use the minified version?

moorepants:
Why commit a non-minified version?

sahilshekhawat:
Let me see if I can compile a minified version too.

sahilshekhawat:
I can, commiting.

moorepants:
Ok, this works for me and solves the kernel restart issue too. Nice work! Here is the screenshot with my browser zoom reset:
selection_094

moorepants:
The play/pause/stop buttons are no longer rounded but that's fine.

sahilshekhawat:
This is what I get on the first runtime. If you run ``scene.display_ipython`` again, it will work fine. I am trying to fix this issue.

sahilshekhawat:
Strangely, minified version is not working properly.

moorepants:
Well, we can commit a non-minified version. I guess it doesn't matter because minifying is only to reduce download speeds and in this case it is downloaded when you install the software. We can go with the un-mini versions.

sahilshekhawat:
I forgot to include it in the header. It is working.

sahilshekhawat:
Its working now, Please review it. Thanks

moorepants:
It all looks good except the slider is still short for me.

moorepants:
selection_095

sahilshekhawat:
Mine is different.
pydy-ipython3

sahilshekhawat:
Can you please check if its updated on your system or not?

moorepants:
The slider is short in Firefox and long in Chromium. Also I can't get the canvas to load in chromium:
selection_096

moorepants:
Chromium is definitely not working for me, here are the JS errors:
selection_097

moorepants:
@sahilshekhawat What browser are you using?

sahilshekhawat:
I am using chrome on linux.

moorepants:
Any idea why chromium is totally failing for me here? It should be the same as chrome, no? I'm also on linux.

sahilshekhawat:
None of the [three.js's examples](http://threejs.org/examples) are working in chromium on linux. So, i guess its a browser issue.

moorepants:
So why the heck would Chrome be different than Chromium???

sahilshekhawat:
Because chromium is open source and some of the bugs might still be there.

moorepants:
Chrome releases are based purely off of stable Chromium releases. The only differences in the browser are some things like a pdf renderer, flash, etc.

For reference my chromium version:

Version 43.0.2357.81 Ubuntu 14.04 (64-bit)

sahilshekhawat:
That's weird!

moorepants:
Ok, I'll look into the Chromimum issue. But the slider is still short in Firefox. Can you test in Firefox? Once that is fixed we can merge this.

moorepants:
What version of Chrome are you running?

sahilshekhawat:
Yeah, I am fixing it right now.

sahilshekhawat:
Its Version 43.0.2357.65 (64-bit)

sahilshekhawat:
Fixed!

moorepants:
The slider is long but this is what I'm seeing now in Firefox:
selection_098

sahilshekhawat:
Its related to zoom in browser. You have to zoom in. The reason is that everything is converted to pixels and all the widths  based on window size.

moorepants:
My browser zoom is at the nominal level.

sahilshekhawat:
Then I guess that notebook has a fixed max-width and thus, a wide screen causes this issue.

moorepants:
Ah, I have it running on a 24" monitor.

sahilshekhawat:
I will try to find a way to fix this. Opening an issue.

moorepants:
Ok, we can merge this as is. Thanks!

moorepants:
I checked off the relevant boxes, leaving the rest as is.

PR #231, opened 2015-06-14
Use Miniconda in Travis for package installation

oliverlee:
Travis will test using Miniconda for installing all required Python
packages in addition to testing with Ubuntu 12.04 site packages for
Numpy, Scipy, and Cython.

  commit message.
- [x] Unit tests have been added for the new feature.
- [x] The PR passes tests both locally (run `nosetests`) and on Travis CI.
- [x] All public methods and classes have docstrings. (We use the [numpydoc
- [x] An explanation has been added to the online documentation. (`docs`
  directory)
- [x] The new feature is documented in the [Release
  follow deprecation cycles.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,21 @@
 language: python
-matrix:
-  include:
-    - python: 2.7
-      env: SYMPY_VERSION="oldest"
-    - python: 2.7
-      env: SYMPY_VERSION="latest"
-    - python: 2.7
-      env: SYMPY_VERSION="master"
-virtualenv:
-  system_site_packages: true
+env:
+  matrix:
+  - SYMPY_VERSION="latest"
+  - SYMPY_VERSION="master"
+  - SYMPY_VERSION="oldest"
+python:
+  - 2.7
 before_install:
   - sudo apt-get update
-  # Remove the version of NumPy pre-installed into the Travis virtualenv so
-  # that we test with the system installed version.
-  - pip uninstall -y numpy
+  - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
+  - chmod +x miniconda.sh
+  - ./miniconda.sh -b
+  - export PATH=/home/travis/miniconda/bin:$PATH
+  - conda update --yes conda
 install:
-  # On Ubuntu 12.04 (Travis's testing machines) numpy is at 1.6.1, SciPy is at
-  # 0.9.0 and Cython is at 0.15.1.
-  - sudo apt-get install -y -qq python-numpy python-scipy cython
-  - sudo apt-get install -y -qq python-coverage
-  - sudo apt-get install -y -qq phantomjs
+  - conda install --yes python=$TRAVIS_PYTHON_VERSION pip numpy scipy cython coverage
+  - sudo apt-get install phantomjs
   # Theano may try to install SciPy from source if the --no-deps flags isn't
   # there.
   - pip install --no-deps Theano==0.6.0


.travis.yml
```diff
@@ -1,24 +1,33 @@
...
-matrix:
-  include:
-    - python: 2.7
-      env: SYMPY_VERSION="oldest"
-    - python: 2.7
-      env: SYMPY_VERSION="latest"
-    - python: 2.7
-      env: SYMPY_VERSION="master"
-virtualenv:
-  system_site_packages: true
```
moorepants: Doesn't this flag have to stay in the file to ensure that the system packages are used? Travis defaults to a virtualenv that doesn't have access to the system packages.

.travis.yml
```diff
@@ -1,24 +1,33 @@
...
+  - "2.7_with_system_site_packages"
 before_install:
-  - sudo apt-get update
+  - sudo apt-get update -qq
   # Remove the version of NumPy pre-installed into the Travis virtualenv so
   # that we test with the system installed version.
-  - pip uninstall -y numpy
+  - pip uninstall -y numpy;
+  - if [[ $TRAVIS_PYTHON_VERSION != "2.7_with_system_site_packages" ]]; then
+      sudo apt-get install -qq libatlas-dev libatlas-base-dev liblapack-dev gfortran;
```
moorepants: I'm not sure any of these are actually needed. Why are they added?

moorepants:
This looks fine except for my two comments.

.travis.yml
oliverlee: Versions of python that are not 2.7 and 3.2 fail in Travis as site packages are not available. This shows an example of what happens if you leave the flag: https://github.com/travis-ci/travis-ci/issues/2219
which also happened with previous versions of this commit: https://travis-ci.org/pydy/pydy/pull_requests

.travis.yml
oliverlee: I thought they were necessary to build scipy but I guess if we install it using conda? I was building it with pip at one point. I'll see if it succeeds in a new build.

moorepants: There are not required for scipy install from apt or conda because those are binary installations. (This is a good reason to switch to conda, because compiling numpy and scipy get old real quick, especially if you use environments or need different versions.)

.travis.yml
moorepants: So travis creates a default virtualenv for Python that doesn't have access to the system python packages. And it also installs a few other packages into this venv (numpy, mock, nose, etc). The process I had here was to first remove the venv numpy because it was a newer version of numpy, and then the venv python will use the system numpy, scipy, etc. which are the versions that shipped with Ubuntu 12.04. If the `system_site_packages` flag isn't set, then I don't think that is what happens. It may be using conda's packages or something in this case.

.travis.yml
```diff
@@ -1,24 +1,33 @@
...
+      wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh;
+      chmod +x miniconda.sh;
+      ./miniconda.sh -b;
+      export PATH=/home/travis/miniconda/bin:$PATH;
```
moorepants: You probably don't want to install conda at all if you want to use the apt packages, like the previous setup. Right here you skip over the virtualenv with this line.

moorepants:
I have a suggestion here. How about we do not use any apt system packages at all anymore. We just use conda for all the testing. The only downside (the reason I hadn't done it yet) is that we will have to bump up our minimum dependencies because anaconda does not ship with all of the versions of numpy, scipy, etc that shipped with Ubuntu 12.04. So we will just bump up the minimum deps for (maybe only) scipy.

.travis.yml
oliverlee: Not necessary: https://travis-ci.org/pydy/pydy/builds/66746826

.travis.yml
oliverlee: Yeah it's skipped.

oliverlee:
Can we install phantomjs with conda?

moorepants:
Ah, still install phantom with apt. All the JS testing stuff needs to be separate from conda as there may not be conda packages for that.

oliverlee:
Only difference from last patch set is removal of comments about package versions in Ubuntu 12.04, as we no longer use those.

Build passes, see here: https://travis-ci.org/pydy/pydy/builds/66777772

moorepants:
+1, merging.

PR #237, opened 2015-06-16
Drop support for IPython < 3.0.

### LINKED ISSUE #235: Drop support for IPython 2
For the 0.3.0 release lets drop support for IPython 2 and make IPython 3 and optional dependency.

moorepants:
Fixes #235.

Also add PR links to some items in the release notes.

I've tested each of the import errors and warnings with no IPython, IPython < 3.0, and IPython 3.0+. Everything works for me locally.

  commit message.
  in the unit test.)
  follow deprecation cycles.)


moorepants:
@pydy/pydy-developers This needs a review for the 0.3.0 release.

pydy/viz/scene.py
```diff
@@ -24,24 +25,20 @@
...
-                            'IPython ' + IPython.__version__ + ' installed. ' +
-                            'IPython related functionalities will not be ' +
-                            'available')
-        ipython_less_than_3 = None
-    elif ipy_ver >= parse_version('2.0') and ipy_ver < parse_version('3.0'):
+    if parse_version(IPython.__version__) < parse_version('3.0'):
+        msg = ('PyDy only supports IPython >= 3.0.0. You have '
+               'IPython {} installed. IPython related functionalities will '
+               'not be available')
+        with warnings.catch_warnings():
```
oliverlee: Don't use catch_warnings().  Instead:

```
warnings.simplefilter('once', ImportWarning)
warnings.warn(msg.format(IPython.__version__), ImportWarning)

**Comment by oliverlee** [comment]
Please change the commit title to:
Drop support for IPython < 3.0

File: pydy/viz/scene.py (hunk 1 above)
Comment (moorepants) [inline pydy/viz/scene.py]: By default Python does not show warnings and I only want these warnings to show once per python session. So if you happen to import this multiple times in single session it will only issue the import warning once.

`catch_warnings` is a nice way to set the filters on a per warning level. I'd also be fine with setting the warning filter to `once` globaly for PyDy. If you have a suggestion of a better pattern for per warning filter control, I can change it.

Comment (oliverlee) [inline pydy/viz/scene.py]: Just use the warning filter with ```once```. The way you have it written was causing the Deprecation Warning to be displayed every single time the method was called when using Python 3.

Comment (moorepants) [inline pydy/viz/scene.py]: So what is the difference in Python 3 about this. I don't understand.

Comment (moorepants) [inline pydy/viz/scene.py]: The reason I used this way of setting the filter was because if you set the warnings filter globally in a module, you'll turn on all warnings for that python session, including, I believe, all the warnings that are normally hidden in any external module you import. Maybe this was a problem when we were testing against NumPy 1.6 and SciPy 0.9.

Comment (moorepants) [inline pydy/viz/scene.py]: You suggestion requires creating sublcassed warnings. I see why you did that in your other PR. I guess I never did that because it seems silly to create a subclass that doesn't have anything overridden. I'll create a `PyDyImportWarning` that subclasses `ImportWarning`.

Comment (oliverlee) [inline pydy/viz/scene.py]: In Python 3, the filter is only applied in the scope of the ```with warnings.catch_warnings()``` block. 

Here's a short example:
```
import warnings

def func():
    with warnings.catch_warnings():
        warnings.filterwarnings('once'):
        warnings.warn('asd', ImportWarning)
        warnings.warn('asd', ImportWarning)

for i in range(2):
    func()
```

If you run with Python 2 and Python 3, it prints the warning a different number of times.

Comment (moorepants) [inline pydy/viz/scene.py]: Thanks, I'll change to what I said above.

**Comment by moorepants** [comment]
Ok, things are fixed.

File: README.rst [hunk 2]
```diff
@@ -349,8 +349,9 @@ Release Notes
 User Facing
 ~~~~~~~~~~~
 
-- Added support Python 3.3 and 3.4. [PR `#38`_]
-- Bumped up the minimum dependencies for NumPy, SciPy, and Cython.
+- Dropped support for IPython < 3.0. [PR `#237`_]
+- Added support Python 3.3 and 3.4. [PR `#229`_]
```
oliverlee: Why is this here? I thought it was already merged with Master.

moorepants: I rebased and then changed it here to point to the PR instead of the original issue to follow the same pattern as the others.

oliverlee: cool thanks

PR #241, opened 2015-06-17
Added shutting down calls to the visualization server

### LINKED ISSUE #232: Clean exit of server
I have few samples of how we can do this: [HERE](https://gist.github.com/sahilshekhawat/a2e79a8df8ce4f31ff4e) and [HERE](https://gist.github.com/sahilshekhawat/93d333af699b82c8bda7)
Both of them fails sometimes because handle_request processes thread in another thread and thus I cannot catch the keyboard interrupt. This fails only when a request is being processing.

sahilshekhawat:
Fixes #232

  commit message.
  in the unit test.)
  follow deprecation cycles.)

### DIFF
--- a/pydy/viz/server.py
+++ b/pydy/viz/server.py
@@ -71,18 +71,10 @@ class Server(object):
     >>> server.run_server()
 
     """
-    def __init__(self, scene_file, directory=None, port=None):
+    def __init__(self, scene_file, directory="static/", port=8000):
         self.scene_file = scene_file
-
-        if directory is None:
-            self.directory = directory
-        else:
-            self.directory = "static/"
-
-        if port is None:
-            self.port = port
-        else:
-            self.port = 8000
+        self.port = port
+        self.directory = directory
 
     def run_server(self):
         # Change dir to static first.
--- a/pydy/viz/server.py
+++ b/pydy/viz/server.py
@@ -103,3 +103,5 @@ class Server(object):
             print("Shutdown confirmed")
             print("Shutting down server...")
             self.httpd.stop()
+        else:
+            print("Resuming operations...")
--- a/pydy/viz/server.py
+++ b/pydy/viz/server.py
@@ -71,7 +71,7 @@ class Server(object):
 
     """
     def __init__(self, scene_file, directory=None, port=None):
-        if scene_file
+        if scene_file:
             self.scene_file = scene_file
 
         if directory:
--- a/pydy/viz/server.py
+++ b/pydy/viz/server.py
@@ -86,10 +86,10 @@ class Server(object):
         print(url)
         webbrowser.open(url)
         print("Hit Ctrl+C to stop the server...")
-        signal.signal(signal.SIGINT, self.stop_server)
+        signal.signal(signal.SIGINT, self._stop_server)
         self.httpd.serve()
 
-    def stop_server(self, signal, frame):
+    def _stop_server(self, signal, frame):
         """
         Confirms and stops the visulisation server
         :param signal:
--- a/pydy/viz/server.py
+++ b/pydy/viz/server.py
@@ -57,9 +57,10 @@ class Server(object):
     Parameters
     ----------
     port : integer
-        Defines the port on which the server will run.
+        Defines the port on which the server will run. If this port is
+        already bind, then it increment 1 until it finds a free port.
     scene_file : name of the scene_file generated for visualization
-        A Valid PyDy generated scene file in 'directory'.
+        A Valid PyDy generated scene file in 'directory' parameter.
     directory : absolute path of a directory
         Absolute path to the directory which contains scene_file with
         all other static files.
@@ -73,12 +74,12 @@ class Server(object):
     def __init__(self, scene_file, directory=None, port=None):
         self.scene_file = scene_file
 
-        if directory:
+        if directory is None:
             self.directory = directory
         else:
             self.directory = "static/"
 
-        if port:
+        if port is None:
             self.port = port
         else:
             self.port = 8000
@@ -121,7 +122,7 @@ class Server(object):
 
         """
         res = raw_input("Shutdown this visualization server ([y]/n)? ")
-        res = res.lower()[0:1]
+        res = res.lower()[0]
         if res == '' or res == 'y':
             print("Shutdown confirmed")
             print("Shutting down server...")
--- a/pydy/viz/server.py
+++ b/pydy/viz/server.py
@@ -4,21 +4,27 @@ import os
 import signal
 import socket
 import webbrowser
-import BaseHTTPServer
-from SimpleHTTPServer import SimpleHTTPRequestHandler
 
+# For python 2 and python 3 compatibility
+if sys.version_info < (3, 0):
+    from SimpleHTTPServer import SimpleHTTPRequestHandler
+    from BaseHTTPServer import HTTPServer
+else:
+    from http.server import SimpleHTTPRequestHandler
+    from http.server import HTTPServer
+    raw_input = input
 
 __all__ = ['Server']
 
 
-class StoppableHTTPServer(BaseHTTPServer.HTTPServer):
+class StoppableHTTPServer(HTTPServer):
     """
     Overrides BaseHTTPServer.HTTPServer to include a stop
     function.
     """
 
     def server_bind(self):
-        BaseHTTPServer.HTTPServer.server_bind(self)
+        HTTPServer.server_bind(self)
         self.socket.settimeout(1)
         self.run = True
 
@@ -49,17 +55,17 @@ class Server(object):
     """
     Parameters
     ----------
-    :param port : integer
+    port : integer
         Defines the port on which the server will run.
-    :param scene_file : name of the scene_file generated for visualization
+    scene_file : name of the scene_file generated for visualization
         Used here to display the url
-    :param directory : path of the directory which contains static and scene files.
+    directory : path of the directory which contains static and scene files.
         Server is started in this directory itself.
 
     Example
     -------
-        >>> server = Server(scene_file=_scene_json_file)
-        >>> server.run_server()
+    >>> server = Server(scene_file=_scene_json_file)
+    >>> server.run_server()
 
     """
     def __init__(self, port=8000, scene_file="Null"):
@@ -97,7 +103,7 @@ class Server(object):
 
         """
         res = raw_input("Shutdown this visualization server ([y]/n)? ")
-        if res is (None or 'y'):
+        if res.lower()[0:1] is (None or 'y'):
             print("Shutdown confirmed")
             print("Shutting down server...")
             self.httpd.stop()
--- a/pydy/viz/server.py
+++ b/pydy/viz/server.py
@@ -1,7 +1,5 @@
 #!/usr/bin/env python
 
-from __future__ import print_function
-
 import os
 import signal
 import socket
--- a/pydy/viz/server.py
+++ b/pydy/viz/server.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import os
+import sys
 import signal
 import socket
 import webbrowser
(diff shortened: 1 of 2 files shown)


sahilshekhawat:
@pydy/pydy-developers This needs a review. Should be a quick one.

pydy/viz/server.py
```diff
@@ -1,37 +1,107 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
```
moorepants: We shouldn't do this. Just make sure to use parentheses on all print statements.

sahilshekhawat: Okay

pydy/viz/server.py
```diff
@@ -1,37 +1,107 @@
...
+                # after setting self.run = False depending on how it
+                # is scheduled.
+                pass
+
+
+class Server(object):
+    """
+    Parameters
+    ----------
+    :param port : integer
```
moorepants: We try to follow the numpydoc style for docstrings, so this would be:

```
Parameters
---------------
port : integer

File: pydy/viz/server.py [hunk 3]
```diff
@@ -1,37 +1,107 @@
...
+        Defines the port on which the server will run.
+    :param scene_file : name of the scene_file generated for visualization
+        Used here to display the url
+    :param directory : path of the directory which contains static and scene files.
+        Server is started in this directory itself.
+
+    Example
+    -------
+        >>> server = Server(scene_file=_scene_json_file)
```
moorepants: Don't indent here.

pydy/viz/server.py
```diff
@@ -1,37 +1,105 @@
...
+    def _stop_server(self, signal, frame):
+        """
+        Confirms and stops the visulisation server
+        :param signal:
+            Required by signal.signal
+        :param frame:
+            Required by signal.signal
+
+        """
+        res = raw_input("Shutdown this visualization server ([y]/n)? ")
```
moorepants: This will fail in Python 3. Oliver had the fix for this in this file before this PR.

pydy/viz/server.py
```diff
@@ -1,37 +1,105 @@
 #!/usr/bin/env python
 
 import os
-import sys
+import signal
+import socket
 import webbrowser
-if sys.version_info < (3, 0):
```
moorepants: This needs to stay in for Python 2/3 compat.

pydy/viz/server.py
```diff
@@ -1,37 +1,105 @@
...
+        """
+        Confirms and stops the visulisation server
+        :param signal:
+            Required by signal.signal
+        :param frame:
+            Required by signal.signal
+
+        """
+        res = raw_input("Shutdown this visualization server ([y]/n)? ")
+        if res is (None or 'y'):
```
oliverlee: Will this eval to ```True``` if the user inputs 'Y'?

moorepants:
Is there any way to have unit tests for this?

pydy/viz/server.py
sahilshekhawat: Nope, I must add it as well

pydy/viz/server.py
sahilshekhawat: I totally missed that. Thanks

pydy/viz/server.py
sahilshekhawat: Done

pydy/viz/server.py
sahilshekhawat: Done

pydy/viz/server.py
moorepants: Maybe make a comment in your PR that says why that is there.

pydy/viz/server.py
sahilshekhawat: What are the options then, should I use ``input()``?

moorepants: See this line: https://github.com/pydy/pydy/blob/master/pydy/viz/scene.py#L26

sahilshekhawat: Go it!

pydy/viz/server.py
```diff
@@ -2,36 +2,112 @@
...
+            Required by signal.signal
+        frame:
+            Required by signal.signal
 
-if __name__ == "__main__":
-    run_server()
+        """
+        res = raw_input("Shutdown this visualization server ([y]/n)? ")
+        res = res.lower()[0:1]
+        if res == '' or res == 'y':
```
sahilshekhawat: This works even for string like "Yes"

pydy/viz/server.py
```diff
@@ -2,36 +2,112 @@
...
+    Example
+    -------
+    >>> server = Server(scene_file=_scene_json_file)
+    >>> server.run_server()
 
+    """
+    def __init__(self, port=8000, scene_file="Null"):
+        self.port = port
+        self.scene_file = scene_file
+        self.directory = "static/"
```
moorepants: Seems like this should be pass in as an argument.

pydy/viz/server.py
```diff
@@ -2,36 +2,112 @@
...
+        Server is started in this directory itself.
 
-__all__ = ['run_server']
+    Example
+    -------
+    >>> server = Server(scene_file=_scene_json_file)
+    >>> server.run_server()
 
+    """
+    def __init__(self, port=8000, scene_file="Null"):
```
moorepants: Why isn't `"Null"` `None`?

pydy/viz/server.py
```diff
@@ -2,36 +2,112 @@
...
+        self.port = port
+        self.scene_file = scene_file
```
moorepants: This should probably be:
```python
if scene_file is not None:
    self.scene_file = scene_file
```

moorepants: My suggestion in the previous commit was to have this call sig:

```python
def __init__(scene_file, directory=None, port=8000)
```
The scene file should be required and the directory can default to the cwd if None is passed in.

sahilshekhawat: I forgot to remove it.

pydy/viz/server.py
```diff
@@ -2,36 +2,112 @@
...
-    httpd = ServerClass(server_address, HandlerClass)
-    sa = httpd.socket.getsockname()
-    print("Serving HTTP on", sa[0], "port", sa[1], "...")
-    print("hit ctrl+c to stop the server..")
-    print("To view visualization, open:\n")
-    url = "http://localhost:"+ str(sa[1]) + "/index.html?load=" + scene_file
-    print(url)
-    webbrowser.open(url)
-    httpd.serve_forever()
+    def run_server(self):
```
moorepants: What happens if port 8000 is taken?

pydy/viz/server.py
sahilshekhawat: Because we need to concatenate it with a string and None will raise an error.

sahilshekhawat: Also, user must know that the there is no file name.

pydy/viz/server.py
sahilshekhawat: I will add that in my new pr on static directory creation. Trying to make it atomic

pydy/viz/server.py
moorepants: `"Null"` is just very anit-python-pattern. I don't like it. `None` should be used and then handled correctly in the code in `__init__`., but my other suggestion is that this should be a required argument.

sahilshekhawat: And I don't know it yet but the url we get from "Null" can be used to give user an option to choose the json files like in ipython notebook.

moorepants: We can worry about that when the feature is implemented. I think scene_file should be a required arg.

pydy/viz/server.py
sahilshekhawat: It throws an error. This needs to be fixed.

pydy/viz/server.py
moorepants: But this is a whole new server code. It seems atomic to add this.

pydy/viz/server.py
sahilshekhawat: Should I change it to None?

sahilshekhawat: or make it a required argument?

moorepants: Seems to me like it is a requirement. What use is the server if I don't have a scene file? I guess you could load it from a url `?scene_file=...` after running the server. If you want to support that use case, then you can make it a kwarg and `=None` or `=`""` but not `="Null"`. I'd prefer `None` as it follows common python patterns.

sahilshekhawat:
Should I create a custom error for empty scene_file?

moorepants:
i don't think that is necessary. if the user passes in a scene_file that isn't a valid valid file name there will be some other errors that raise, no?

sahilshekhawat:
it is for an empty scene file name.

moorepants:
What do you mean by empty?

sahilshekhawat:
When user does not pass the scene_file to the server. We should raise an error, no?

moorepants:
If your docstring says:
```
scene_file : string
    A valid path to a file in the ``directory``.
```
Then you can use the "garbage in, garbage out" rule.

moorepants:
If the user doesn't pass in a required argument they will get an error from Python that says they are missing an argument.

pydy/viz/server.py
```diff
@@ -2,36 +2,133 @@
...
-    print("hit ctrl+c to stop the server..")
-    print("To view visualization, open:\n")
-    url = "http://localhost:"+ str(sa[1]) + "/index.html?load=" + scene_file
-    print(url)
-    webbrowser.open(url)
-    httpd.serve_forever()
+    def _check_port(self, port):
+        soc = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        result = soc.connect_ex(('127.0.0.1', port))
+        if result == 0:
```
oliverlee: return result == 0

sahilshekhawat: Fixed.

sahilshekhawat:
@oliverlee Can you please review this PR. It needs to be merged before my other PR can be reviewed.

pydy/viz/server.py
```diff
@@ -2,36 +2,129 @@
...
+        signal:
+            Required by signal.signal
+        frame:
+            Required by signal.signal
 
-if __name__ == "__main__":
-    run_server()
+        """
+        res = raw_input("Shutdown this visualization server ([y]/n)? ")
+        res = res.lower()[0:1]
```
oliverlee: res.lower()[0] can be used instead.

sahilshekhawat: Yup! changing.

pydy/viz/server.py
```diff
@@ -2,36 +2,129 @@
...
+                # is scheduled.
+                pass
+
+
+class Server(object):
+    """
+    Parameters
+    ----------
+    port : integer
+        Defines the port on which the server will run.
```
oliverlee: This isn't the case if the port is taken.

pydy/viz/server.py
```diff
@@ -2,36 +2,129 @@
...
+    Example
+    -------
+    >>> server = Server(scene_file=_scene_json_file)
+    >>> server.run_server()
+
+    """
+    def __init__(self, scene_file, directory=None, port=None):
+        self.scene_file = scene_file
 
+        if directory:
```
oliverlee: Comparisons to singletons like None should always be done with is or is not:
``` if directory is not None ```

**Comment by sahilshekhawat** [comment]
I have addressed all the concerns here. It would be very helpful if someone can please review it. I think it can be merged. Thanks

File: pydy/viz/server.py [hunk 16]
```diff
@@ -2,36 +2,130 @@
...
+    def __init__(self, scene_file, directory=None, port=None):
+        self.scene_file = scene_file
 
+        if directory is None:
+            self.directory = directory
+        else:
+            self.directory = "static/"
 
-__all__ = ['run_server']
+        if port is None:
```
moorepants: If you have a default value for `port` and `directory` you should just set it in the call signature:

```
def __init__(self, scene_file, directory='static', port=8000):
```

Sorry if I've confused you about this.

moorepants: But, I don't see how using `directory='static'` is a good idea. That's why I suggested using `directory=None` and then if that was true setting `directory = os.getcwd()`.

sahilshekhawat: Okay , I am updating it.
Static directory gets created in current working directory itself.

moorepants:
I'm merging this without tests. I'll raise an issue to fix that.

PR #242, opened 2015-06-18
Pr240 fixing static directory creation issue

sahilshekhawat:
  commit message.
  in the unit test.)
  follow deprecation cycles.)

### DIFF
--- a/pydy/viz/scene.py
+++ b/pydy/viz/scene.py
@@ -185,7 +185,7 @@ class Scene(object):
     @system.setter
     def system(self, new_system):
 
-        if new_system is not None and not isinstance(new_system, System):
+        if not isinstance(new_system, System):
             msg = "{} should be a valid pydy.System object".format(new_system)
             raise TypeError(msg)
 
@@ -232,7 +232,7 @@ class Scene(object):
         if new_time is None:
             self._time = new_time
         else:
-            self._time = np.asarray(new_time)
+            self._time = np.array(new_time)
 
     @property
     def states_symbols(self):
--- a/pydy/viz/scene.py
+++ b/pydy/viz/scene.py
@@ -3,6 +3,7 @@
 # standard library
 from __future__ import division
 import os
+import sys
 import warnings
 import json
 import distutils
@@ -12,6 +13,7 @@ from collections import OrderedDict
 from pkg_resources import parse_version
 
 # external
+import numpy as np
 from sympy import latex
 from sympy.physics.mechanics import ReferenceFrame, Point
 
@@ -20,16 +22,15 @@ from .camera import PerspectiveCamera
 from .server import run_server
 from .light import PointLight
 from ..system import System
-from ..utils import PyDyImportWarning
+from ..utils import PyDyImportWarning, PyDyDeprecationWarning
 
-import sys
 if sys.version_info > (3, 0):
     raw_input = input
-
 
 __all__ = ['Scene']
 
 warnings.simplefilter('once', PyDyImportWarning)
+warnings.simplefilter('once', PyDyDeprecationWarning)
 
 try:
     import IPython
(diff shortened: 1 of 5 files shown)


sahilshekhawat:
Please review #241 first so that I can merge it with new server and test it before merging.

sahilshekhawat:
There is one issue now that we are not keeping previous scene files. When we  run all cells again in Ipython notebook, notebook checks that the code has not been changed and does not runs it again, as a result there is no new visualization file while the old one has been deleted. This causes scene.display_ipython() to raise error.

pydy/viz/scene.py
```diff
@@ -154,6 +178,146 @@ def reference_frame(self, new_reference_frame):
...
             self._reference_frame = new_reference_frame
 
+    @property
+    def system(self):
+        return self._system
+
+    @system.setter
+    def system(self, new_system):
+
+        if new_system is not None and not isinstance(new_system, System):
```
oliverlee: only need to check isinstance

sahilshekhawat: Acknowledged.

pydy/viz/scene.py
```diff
@@ -154,6 +178,146 @@ def reference_frame(self, new_reference_frame):
...
+                    msg = ('The time array length, {}, does not match the '
+                           'length of the state trajectories array, {}.')
+                    raise ValueError(msg.format(len(new_time), len_traj))
+        except AttributeError:
+            pass
+
+        if new_time is None:
+            self._time = new_time
+        else:
+            self._time = np.asarray(new_time)
```
oliverlee: Shouldn't this be np.array?

sahilshekhawat: hmm, yes this needs to be np.array.

pydy/viz/scene.py
```diff
@@ -460,9 +595,37 @@ def remove_static_html(self, force=False):
...
+
+        if not silent:
+            print("Copying Simulation data.")
+        try:
+            shutil.move(os.path.join(os.getcwd(), self._scene_json_file),
+                        os.path.join(self.static_url, self._scene_json_file))
+            shutil.move(os.path.join(os.getcwd(), self._simulation_json_file),
+                        os.path.join(self.static_url, self._simulation_json_file))
+        except IOError:
+            pass
```
oliverlee: Should this really pass if the file move fails?

sahilshekhawat: No, but right now I have an issue with ipython notebook. When we run notebook seconds time without reloading, no visualization file is created and this needs to be fixed. This will be fixed before merging.

pydy/viz/scene.py
moorepants: I had it that way at first and the tests errored. On construction of the instance system maybe be set to None, so this will error for no reason. Passing in None or a valid System are the two valid options.

sahilshekhawat: I have changed it back to this  as test with system=None were failing.

pydy/viz/scene.py
```diff
@@ -48,71 +50,93 @@
...
+            An ordered sequence of the SymPy functions that represent the
+            states. The order must match the order of the
+            ``states_trajectories``.
+        states_trajectories : array_like, shape(n, m), optional, default=None
+            A two dimensional array with numerical values for each state at
+            each poitn in time during the animation.
+
+        Notes
+        =====
+        The user is allow to supply either system or time, constants,
```
moorepants: allow > allowed

pydy/viz/scene.py
```diff
@@ -154,6 +178,146 @@ def reference_frame(self, new_reference_frame):
...
+                try:
+                    if getattr(self, attr) is not None:
+                        raise ValueError(msg.format(attr))
+                except AttributeError:
+                    pass
+
+        self._system = new_system
+
+    @property
+    def time(self):
```
moorepants: Let's change this attribute to `times` to match the System class.

pydy/viz/scene.py
```diff
@@ -460,9 +595,37 @@ def remove_static_html(self, force=False):
         else:
             print("aborted!")
 
+    def copy_resources(self, silent=False):
```
moorepants: I think this should be a private method, leading `_`.

pydy/viz/scene.py
```diff
@@ -460,9 +595,37 @@ def remove_static_html(self, force=False):
...
+        """
+        Copies simulation and static files to a directory pointed
+        by self.static_url attribute.
+        It overwrites if such directory is already present.
+        """
+        dst = self.static_url
```
moorepants: `static_url` should be private as well.

pydy/viz/scene.py
```diff
@@ -460,9 +595,37 @@ def remove_static_html(self, force=False):
...
+                        os.path.join(self.static_url, self._scene_json_file))
+            shutil.move(os.path.join(os.getcwd(), self._simulation_json_file),
+                        os.path.join(self.static_url, self._simulation_json_file))
+        except IOError:
+            pass
+
     def display(self):
         """Displays the scene in the default webbrowser."""
-        self.create_static_html()
+        self.copy_resources()
```
moorepants: I was imaginging this working a little differently.

```python
def display(self):
    hidden_dir = ...
    self._generate_json(directory=hidden_dir)
    run_server(self._scene_json_file, directory=hidden_dir)
```

moorepants: I don't think `copy_resources` is necessary. `_generate_json` allows you to create the files in the directory that you desire. I think that may fix your ipython rerun issue too.

moorepants: the `create_static_html` method should also use this design. It should copy over the files in the cwd directory `static` and then call `_generate_json` and place those files in the directory. Then the `static` dir contains the whole "app" for that one simulation that the user may want to hold on to.

sahilshekhawat: Yes, this makes sense now and I think the bug will be solve after this. On it.

pydy/viz/scene.py
sahilshekhawat: Fixed.

sahilshekhawat:
Yes, that fixed the issue with ipython. Now it is working fine. A directory pydy-resources is created in current directory which consists of json files and static data.

moorepants:
Do you think this would work for cross platform hidden directories:

http://stackoverflow.com/questions/25432139/python-cross-platform-hidden-file

pydy/viz/scene.py
```diff
@@ -462,8 +605,8 @@ def remove_static_html(self, force=False):
 
     def display(self):
         """Displays the scene in the default webbrowser."""
-        self.create_static_html()
-        server = Server(scene_file=self._scene_json_file)
+        self._generate_json()
+        server = Server(scene_file=self._scene_json_file, directory=self._static_url)
```
moorepants: It isn't clear to me when `self._static_url` is created.

sahilshekhawat: I am assigning it here: https://github.com/pydy/pydy/pull/242/files#diff-b43b925b1507940800c850f885764b51R139

sahilshekhawat:
Normal visualizations can work for cross-platform hidden directories but again Ipython notebook will not work.

pydy/viz/scene.py
```diff
@@ -48,71 +50,93 @@
...
+                          'time': None,
+                          'constants': None,
+                          'states_symbols': None,
+                          'states_trajectories': None,
+                          'frames_per_second': 30}
+        default_kwargs.update(kwargs)
+        for k, v in default_kwargs.items():
+            setattr(self, k, v)
+
+        self._static_url = os.path.join(os.getcwd(), "pydy-resources")
```
moorepants: I'm not sure that this is the best place to create the path to this dir. The user coudl theorectically change directory in between the calls to `__init__()` and `display()`.

I would make a class level attribute called `pydy_dir` or something. And then in the `display()` methods you should do this os.path.join.

pydy/viz/scene.py
```diff
@@ -350,16 +497,15 @@ def generate_visualization_json_system(self, system, **kwargs):
...
+        self.system = system
+        prefix = None
+        for k, v in kwargs.items():
+            if k == 'fps':
+                self.frames_per_second = v
+            if k == 'outfile_prefix':
+                prefix = v
+        self._generate_json(prefix=prefix)
 
     def create_static_html(self, overwrite=False, silent=False):
```
moorepants: `create_static_html` should create a directory called `static` in the cwd and fill it with both the files from the `static` directory in the pydy source tree and the two json files for the current simulation (i.e. whatever the simulation is at the time `create_static_html` is called. Then the user has a single directory that they could copy to their website server and display the visualization online for others to view. This is the method that we can use to embed everything in the pydy website eventually.

moorepants: There definitely need to be a call to `_generate_json()` in this method, otherwise there are no `self._scene_json_file`, etc defined.

pydy/viz/scene.py
moorepants: Found it in `__init__`, see my note there.

pydy/viz/scene.py
```diff
@@ -462,8 +605,8 @@ def remove_static_html(self, force=False):
 
     def display(self):
         """Displays the scene in the default webbrowser."""
-        self.create_static_html()
-        server = Server(scene_file=self._scene_json_file)
+        self._generate_json()
```
moorepants: Don't you have to supply the directory kwarg here, other wise the json files will not be copied into the `_static_url` directory.

pydy/viz/scene.py
moorepants: And why is this called a `url`?

moorepants:
> Normal visualizations can work for cross-platform hidden directories but again Ipython notebook will not work.

Got it. Thanks.

oliverlee:
Your patch fails linting requirements. Check the Travis build log for details.

moorepants:
Closing in favor of #245.

PR #247, opened 2015-06-23
[WIP] Added using_joints classmethod to pydy.system to create system using joints.

sahilshekhawat:
Refer to https://github.com/sympy/sympy/pull/9559 for changes to rigidbody.

### DIFF
--- a/pydy/tests/test_system.py
+++ b/pydy/tests/test_system.py
@@ -3,9 +3,10 @@
 import numpy as np
 from numpy import testing
 import sympy as sm
-from sympy.physics.mechanics import dynamicsymbols
+from sympy.physics.mechanics import dynamicsymbols, Point, ReferenceFrame
 from scipy.integrate import odeint
 
+from ..bodies import Ground
 from ..system import System
 from ..models import multi_mass_spring_damper, n_link_pendulum_on_cart
 from ..utils import sympy_equal_to_or_newer_than
@@ -37,6 +38,7 @@ class TestSystem():
         # -----------------------------------
         sys = System(self.kane)
 
+        assert sys.is_using_joints is False
         assert (sys.constants_symbols ==
                 set(sm.symbols('k0, m0, g, c0')))
         assert sys.specifieds_symbols == {self.specified_symbol}
@@ -395,3 +397,27 @@ class TestSystem():
         sys = System(self.kane, times=times)
         with testing.assert_raises(NotImplementedError):
             sys.generate_ode_function(generator='made-up')
+
+
+class TestJointSystem():
+
+    def setup(self):
+        self.system = System.using_joints()
+
+    def test_init(self):
+        assert isinstance(self.system.ground, Ground)
+        assert isinstance(self.system.origin, Point)
+        assert isinstance(self.system.reference_frame, ReferenceFrame)
+        assert self.system.bodies == list()
+        assert self.system.is_using_joints is True
+
+    def test_gravity(self):
+        assert self.system.gravitational_constant == sm.Symbol('gravity')
+        assert self.system.gravity == sm.Symbol('gravity') * \
+                                      self.system.reference_frame.z
+
+    def test_generate_eoms(self):
+        # TODO
+
+    def test_visualize(self):
+        # TODO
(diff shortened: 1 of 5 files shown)


pydy/tests/test_system.py
```diff
@@ -395,3 +397,27 @@ def test_integrate(self):
         sys = System(self.kane, times=times)
         with testing.assert_raises(NotImplementedError):
             sys.generate_ode_function(generator='made-up')
+
+
+class TestJointSystem():
+
+    def setup(self):
+        self.system = System.using_joints()
```
moorepants: This is interesting. I don't know if I've even seen this done in Python. I think the typical way to do this would to support an `__init__` method that took no arguments. Or subclass System to make a JointSystem which overrides the `__init__` method.

moorepants:
@sahilshekhawat Please focus only on the unit tests as we discussed. It is better to spend time writing the units tests so we can reveiew the design, which will ask for changes to the unit tests. If you write the unit tests and the implementation code, then when we ask for a design change in the review, you'll have twice and much work to do in the edit. At most write out "empty" classes/methods in the implementation, i.e. just write the class names and methods names with call signatures and what they will return if you need that.

pydy/tests/test_system.py
sahilshekhawat: http://stackoverflow.com/questions/19305296/multiple-constructors-in-python-using-inheritance
Its my reference

> A simple example from the standard library is datetime.datetime, which can be constructed with now, fromtimestamp, or a few other alternate constructors, besides the default.

Also, I want to keep our present classes intact as long as we have options. Wanted to get your views on this.

sahilshekhawat:
Yeah, sure! I am working on unit tests now, its just that tarun asked me to work on some code too to discuss some basic problems.

pydy/tests/test_system.py
sahilshekhawat: Else we will need to do major restructuring which might induce many bugs too.

moorepants: It is just an odd pattern that typically Python users will be unfamiliar with. I like the other three options better than the first in that stackoverflow answer. I do not think that the other more typically solutions will introduce an unnecessary amount of bugs. If we want one system class, no sublcasses, then simply make all arguments to System be kwargs instead of args. If you get `method=None` then you instantiate it without a method. The user can then build the system and once ready, it will be able to generate the needed Method class behind the scenes.

moorepants:
Ok, if Tarun wanted to see some of that code, that's fine.

moorepants:
My suggestion about writing only unit tests, is from an efficiency standpoint. If you present the design through unit tests for us to review then when we ask fro changes you only have to change the unit tests instead of changing both the unit tests and the implementation. That way when you finally write the implementation, you only have to write it once.

@tarzzz Let me know if you want @sahilshekhawat to present things in other ways so we aren't requesting two different things from him.

pydy/tests/test_system.py
```diff
@@ -395,3 +397,27 @@ def test_integrate(self):
...
+
+    def test_init(self):
+        assert isinstance(self.system.ground, Ground)
+        assert isinstance(self.system.origin, Point)
+        assert isinstance(self.system.reference_frame, ReferenceFrame)
+        assert self.system.bodies == list()
+        assert self.system.is_using_joints is True
+
+    def test_gravity(self):
+        assert self.system.gravitational_constant == sm.Symbol('gravity')
```
moorepants: The user should be able to optionally set this symbol and it should also be able to be set to zero. I'd make the default simply `'g'`.

tarzzz:
@moorepants @sahilshekhawat I wanted Sahil to create a layout of Class methods in a wiki page, for better understanding. For This PR, I would suggest him to keep only the test cases.

sahilshekhawat:
I have created a layout with funtions name and doc string in the code itself. Should i remove it from there and add to wiki?

tarzzz:
You should keep this PR with only unit tests. You can keep the code in a separate branch, and remove from here for now. Please do create a wiki page with a brief design layout as well.

sahilshekhawat:
Got it, changing!

tarzzz:
Thanks

moorepants:
If this is abandoned now, please close.

PR #249, opened 2015-06-26
[WIP] Added basic unit tests for system, bodies, joints and utils.

sahilshekhawat:
I am writing doc strings of functions here: https://github.com/pydy/pydy/wiki/System-Joints-API-GSoC-2915

### DIFF
--- a/pydy/system.py
+++ b/pydy/system.py
@@ -610,3 +610,42 @@ class System(object):
                 *self._Kane_inlist_insyms()))
         constants.remove(dynamicsymbols._t)
         return constants
+
+    def add_gravity(self, direction=None):
+        """Assigns gravity, by default in the direction of reference_frame.z
+        """
+        if direction is None:
+            direction = self.reference_frame.z
+        self._apply_gravity()
+        # TODO
+
+    def _apply_gravity(self):
+        """Applies gravity force to each body in body_list. Called after calling
+        add_gravity and also when a new body is added."""
+        # TODO
+
+    def add_body(self, body):
+        """Adds body to the body_list, assigns system to the body, gravity. Only
+        called when body is not connected to any joints. If it is create a joint
+        first and then add joints to the system
+        """
+        # TODO
+
+    def add_joint(self, joint):
+        """Adds joints to the system and body involved in creating the joints to
+        the system
+        """
+        # TODO
+
+    def get_eom(self):
+        """Generates the equations of motion, it iterates over all parent-body tree
+        and create a tree structure for their center of masses and reference frames
+        """
+        # TODO
+
+    def visualize(self, ipython=True):
+        """
+        Calculates the EoMs and using the data from add_decoration from bodies
+        calls display or display_ipython and starts the server.
+        """
+        # TODO
--- a/pydy/tests/test_bodies.py
+++ b/pydy/tests/test_bodies.py
@@ -1,0 +1,23 @@
+#!/usr/bin/env python
+
+SYMPY_VERSION = sm.__version__
+
+
+class TestBodies():
+
+    def test_body_init(self):
+        body = Body('body')
+
+        assert body.name is None
+        assert body.mass == Symbol('mass')
+
+        sys = System(method='joints')
+        sys.add_body(body)
+
+        assert body.name == 'body_1'
+
+        body_length = Length(body.name)
+        assert body_length.symbol == Symbol(body.name + "_length")
+
+        body_length_value = Length(body.name, 2)
+        assert body_length_value.value == 2
--- a/pydy/tests/test_joints.py
+++ b/pydy/tests/test_joints.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+
+SYMPY_VERSION = sm.__version__
+
+from ..bodies import Body
+from ..joints import PinJoint
+from ..system import System
+
+
+class TestPinJoint():
+    def setup(self):
+        self.sys = System(method='joints')
+
+    def test_joints_parent_child_relationship(self):
+        body = Body('body')
+        pin_joint = PinJoint(self.ground, 0, body, sys.reference_frame.x * Symbol('l'))
+
+        assert self.sys.ground.child is None
+        assert body.parent is None
+
+        self.sys.add_joint(pin_joint)
+
+        assert self.sys.ground.child == body
+        assert body.parent == self.sys.ground
+
+    def test_joints_system_body_list(self):
+        body = Body('body')
+        pin_joint = PinJoint(self.ground, 0, body, sys.reference_frame.x * Symbol('l'))
+
+        assert self.sys.body_list == list()
+        self.sys.add_joint(pin_joint)
+        assert body in self.sys.body_list
+
+    def test_joint_locate_pin_point(self):
+        # TODO
+
--- a/pydy/tests/test_system.py
+++ b/pydy/tests/test_system.py
@@ -3,12 +3,14 @@
 import numpy as np
 from numpy import testing
 import sympy as sm
+from sympy.physics.vector import ReferenceFrame
 from sympy.physics.mechanics import dynamicsymbols
 from scipy.integrate import odeint
 
+from ..bodies import Ground, Body
 from ..system import System
 from ..models import multi_mass_spring_damper, n_link_pendulum_on_cart
-from ..utils import sympy_equal_to_or_newer_than
+from ..utils import sympy_equal_to_or_newer_than, AssignSymbol
 
 SYMPY_VERSION = sm.__version__
 
@@ -395,3 +397,67 @@ class TestSystem():
         sys = System(self.kane, times=times)
         with testing.assert_raises(NotImplementedError):
             sys.generate_ode_function(generator='made-up')
+
+    def test_joint_system_init(self):
+        sys = System(method='joints')
+
+        # test that none of attributes set when system is called with kane
+        # are set.
+        assert hasattr(sys, 'method')
+        assert sys.method == 'joints'
+        assert sys.constants == dict()
+        assert sys.specifieds == dict()
+        assert sys.initial_conditions == dict()
+
+        # test all the attributes used to build system using joints are set.
+        assert isinstance(sys.ground, Ground)
+        assert sys.origin == sys.ground.origin
+        assert sys.gravity is None
+        assert isinstance(sys.reference_frame, ReferenceFrame)
+        assert sys._body_list == []
+
+    def test_joint_system_gravity(self):
+        sys = System(method='joints')
+        sys.add_gravity()
+
+        assert sys.gravitational_constant == Symbol('g')
+        assert sys.gravity == sys.gravitatinal_constant * sys.reference_frame.z
+
+    def test_joint_system_add_body(self):
+        sys = System(method='joints')
+        body = Body('Body')
+
+        sys.add_body(body)
+
+        assert sys.body_list == [body]
+        assert body.system == sys
+        assert body.parent is None
+
+    def test_joint_system_add_gravity(self):
+        sys = System(method='joints')
+        body = Body('Body')
+        sys.add_gravity()
+        gravity_force = body.mass * sys.gravitational_constant * sys.reference_frame.z
+
+        assert body.force_list == list()
+
+        sys.add_body(body)
+
+        assert gravity_force in body.force_list
+
+    # add_joint tested in test_joints
+    def test_system_assign_name(self):
+        system = System(method='joints')
+        body = Body()
+        assert body.name is None
+
+        system.add_body(body)
+        assign_symbol = AssignSymbol()
+
+        assert body.name == assign_symbol.get_body_name()
+        assert body.name == 'body_1'
+        assert isinstance(system.assign_symbol, AssignSymbol)
+
+    def test_system_visualize(self):
+        # TODO tests calling display or display_ipython and then starting server
+        # after assigning
--- a/pydy/tests/test_utils.py
+++ b/pydy/tests/test_utils.py
@@ -1,10 +1,14 @@
 #!/usr/bin/env python
 
+from sympy import Symbol
+
 from pkg_resources import parse_version
 from setuptools import __version__ as SETUPTOOLS_VERSION
 from nose.tools import assert_raises
 
-from ..utils import sympy_equal_to_or_newer_than
+from ..bodies import Body
+from ..utils import sympy_equal_to_or_newer_than, Mass, Length, \
+        GravitationalConstant, AssignSymbol
 
 
 def test_sympy_equal_to_or_newer_than():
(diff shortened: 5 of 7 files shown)


pydy/tests/test_utils.py
```diff
@@ -18,3 +22,61 @@ def test_sympy_equal_to_or_newer_than():
     if parse_version(SETUPTOOLS_VERSION) >= parse_version('8.0'):
         with assert_raises(ValueError):
             sympy_equal_to_or_newer_than('0.7.7', '0.7.6-git')
+
+def test_mass():
+
+    mass = Mass()
+    assert mass.symbol == Symbol('m')
+    assert mass.value == 1
```
moorepants: How is this attribute used? Everything you are designing should be able to output fully symbolic eoms. Our current method is to pass in numerical values to system as a `sys.constants = {}`.

sahilshekhawat: Its my attempt to provide default values and user have the option to change those initial conditions. mass.value will be used in system.initial_conditions

pydy/tests/test_utils.py
```diff
@@ -1,10 +1,14 @@
 #!/usr/bin/env python
 
+from sympy import Symbol
+
```
moorepants: This whole addition seems like you are adding classes that 1) could be namedtuples or 2) are likely unnecessary. What is the reason to add all of these?

If you go this route, there are an infinite number of class types that can be added. I could possibly imagine that a class called `Constant` or `Parameter` could possibly be useful, but it still seems like we'd be adding lots of frivolous classes here.

From the "symbolic side" all you need is a SymPy symbol for each model parameter/constant. Eventually you want to tie a numerical value to that, but be able to easily change that numerical value for different simulations and analyses. Write now our rhs functions are generated as "parameter value agnostic" and the user can provide them at the time of evaluation of the EoMs, so I don't really see how you will propagate these values to the user when they simply want to evaluate the EoMs.

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,23 @@
...
+    def test_body_init(self):
+        body = Body('body')
+
+        assert body.name is None
+        assert body.mass == Symbol('mass')
+
+        sys = System(method='joints')
+        sys.add_body(body)
+
+        assert body.name == 'body_1'
```
moorepants: So why does the body change it's name if you add it to a System? What happens if I set a name for the body and then add it to the system? I would hope that the name I set is retained.

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python
+
+SYMPY_VERSION = sm.__version__
+
+
+class TestBodies():
+
+    def test_body_init(self):
+        body = Body('body')
```
moorepants: I though that `'body'` is the name of the body, yet in line 11 it is set to None. What is this required arg for?

pydy/tests/test_bodies.py
sahilshekhawat: The idea is that if user does not passes a name, it automatically increments and creates a unique body name which is required to define other attributes of the body.

pydy/tests/test_bodies.py
sahilshekhawat: Its is optional to pass the name.

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,23 @@
...
+        assert body.name is None
+        assert body.mass == Symbol('mass')
+
+        sys = System(method='joints')
+        sys.add_body(body)
+
+        assert body.name == 'body_1'
+
+        body_length = Length(body.name)
+        assert body_length.symbol == Symbol(body.name + "_length")
```
moorepants: If the user passes a string to `Length` why do you modify that string and append `_length` here?

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,36 @@
...
+from ..system import System
+
+
+class TestPinJoint():
+    def setup(self):
+        self.sys = System(method='joints')
+
+    def test_joints_parent_child_relationship(self):
+        body = Body('body')
+        pin_joint = PinJoint(self.ground, 0, body, sys.reference_frame.x * Symbol('l'))
```
moorepants: Is there any significance to passing `PinJoint` `sys.reference_frame.x * Symbol('l')` instead of `some_reference_frame.x * Symbol('l')`?

pydy/tests/test_bodies.py
sahilshekhawat: I see your point, I should not be adding all thses redundant classes. Lets just stick to sympy.Symbol() and let user pass them up.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,36 @@
...
+
+        assert self.sys.ground.child is None
+        assert body.parent is None
+
+        self.sys.add_joint(pin_joint)
+
+        assert self.sys.ground.child == body
+        assert body.parent == self.sys.ground
```
moorepants: So sympy.physics.vector.ReferenceFrame and sympy.physics.vector.Point both have tree topologies already build in. How will this new tree topology interact with the previous ones?

pydy/tests/test_joints.py
sahilshekhawat: User can pass any reference_frame provided there is a relation between the reference frame and the body.

pydy/tests/test_joints.py
sahilshekhawat: system will handle them. Lets suppose that we added a Joint with a parent and child. Assuming that the joint is PinJoint, it defined an angular velocity of chld w.r.t parent. Now, we we will generate EoMs, system will iterate through all the parent and child and creating the tree topology of the  Points and reference_frames.

pydy/tests/test_system.py
```diff
@@ -395,3 +397,67 @@ def test_integrate(self):
         sys = System(self.kane, times=times)
         with testing.assert_raises(NotImplementedError):
             sys.generate_ode_function(generator='made-up')
+
+    def test_joint_system_init(self):
+        sys = System(method='joints')
```
moorepants: This made me think of an idea. So we already have `Methods` classes. A method essentially takes a collection of RigidBodies, Particles, and Forces/Torques and outputs the symbolic equations of motion. The RigidBodies and Particles already have their kinematics prescribed before they arrive to the Method class. You're new system construction stuff could be considered a Method too. Except your your Method requires, Bodies, Particles, Forces/Torques, and Joints where the kinematics are not prescribed for the RigidBodies and Particles, but instead you pass in Joints. But inside your Method, you first use the Joints to prescribe the kinematics and then use another Method to finally compute the symbolic EoMs.

So the point I'm getting at is that you could create a new Method class called `JointMethod` or something that housed the logic of prescribing the kinematics to the RigidBodies and Particles. This would mean that System doesn't really have to change. We would still initialize a System with a method class.

This is a derailment from the current ideas you've developed, but maybe one worth pondering. It just occurred to me.

pydy/tests/test_system.py
```diff
@@ -395,3 +397,67 @@ def test_integrate(self):
...
+        # test all the attributes used to build system using joints are set.
+        assert isinstance(sys.ground, Ground)
+        assert sys.origin == sys.ground.origin
+        assert sys.gravity is None
+        assert isinstance(sys.reference_frame, ReferenceFrame)
+        assert sys._body_list == []
+
+    def test_joint_system_gravity(self):
+        sys = System(method='joints')
+        sys.add_gravity()
```
moorepants: I should probably be an option for the user to set the symbol used for gravity.

pydy/tests/test_system.py
```diff
@@ -395,3 +397,67 @@ def test_integrate(self):
...
+        assert gravity_force in body.force_list
+
+    # add_joint tested in test_joints
+    def test_system_assign_name(self):
+        system = System(method='joints')
+        body = Body()
+        assert body.name is None
+
+        system.add_body(body)
+        assign_symbol = AssignSymbol()
```
moorepants: It isn't clear what an `AssignSymbol()` class is.

pydy/tests/test_system.py
sahilshekhawat: Yup! I am removing all the redundant classed from utils and user can pass the symbol the gravity.

pydy/tests/test_bodies.py
moorepants: Ok, so you will need a test to make sure it doesn't override the user's preferred choice.

pydy/tests/test_bodies.py
moorepants: So shouldn't the name not be None in the test below.

pydy/tests/test_bodies.py
sahilshekhawat: Okay

pydy/tests/test_bodies.py
moorepants: Think some about a Parameter class though. We have an infinite number of parameters and they all have a symbol and potentially a value. So there could possible be need for that, but it could just be a tuple or a namedtuple. I'm not sure it is that valuable to have a Parameter class but I could be convinced if you have a utility in mind.

pydy/tests/test_joints.py
moorepants: Ok.

pydy/tests/test_joints.py
moorepants: So this is a new tree topology that somehow combines a Point and a ReferenceFrame?

moorepants:
@sahilshekhawat These are great! It really is clear what you are trying to do. Please continue with this. The more you can get fleshed out here the better. It is very easy to read and comment on the design when it is laid out like this. What do you think?

pydy/tests/test_utils.py
moorepants: Keep in mind that what you build should work with only symbolics provided. Ideally a user can fully describe the bodies/particles, the joints, and the forces with only symbolic parameters and it will ultimately spit out symbolic equations of motion. We currently have a dividing line between symbolics and numerics. All of our EoM generation methods should work in a purely symbolic way.

sahilshekhawat:
Although, it takes a lot of time to think without the code for the previous tests but I agree that its the moost efficient way to do things as iterating it is very easy.

moorepants:
The wiki page is nice too, but why not simply commit those "empty" class definitions here. You can create unit tests and "empty" classes with the docstrings all in this PR.

moorepants:
Or make a separate PR with the empty class definitions.

pydy/tests/test_system.py
sahilshekhawat: This class is the one which assigns a unique name when no name is supplied. It provides unique names for anything like Body, Joints etc.

moorepants: Ok, well create a unit test for it so that we can see how it works.

sahilshekhawat:
Yeah, let me commit them too. I just put them on hold, thought that I am expected to push only unit tests.

pydy/tests/test_system.py
sahilshekhawat: here: https://github.com/pydy/pydy/pull/249/files#diff-1ae069562fff981b26a9234936f0a3beR74

moorepants:
I think it is a good idea to submit empty class/function defs and unit tests in tandem. They once we all agree with the design, you can fill in the empty classes and we can merge them.

But let's also think about all this in terms of atomic commits. Can you make a PR for each new class? The PR can contain unit tests and an empty class def. Then as we agree on each class, we can merge them in order. If you try to do all of these classes in one PR, we are going to have one huge massive PR that is going to be very hard to merge.

You can continue like you are in this one PR for now to flesh things out more, but you will need to move to more atomic PRs (probably sooner than later).

What do you think?

sahilshekhawat:
Atomic PRs makes so much sense now. They will speed up the things.

pydy/system.py
```diff
@@ -610,3 +610,42 @@ def _Kane_constant_symbols(self):
                 *self._Kane_inlist_insyms()))
         constants.remove(dynamicsymbols._t)
         return constants
+
+    def add_gravity(self, direction=None):
```
tarzzz: Instead of add_gravity, it should be more generalized to *add and apply* different forces, and torques(if possible) in a single generic ClassMethod

sahilshekhawat: We will include that too within a separate ``Force`` class. Gravity is for
a quick addition for something used almost every time.

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,23 @@
...
+SYMPY_VERSION = sm.__version__
+
+
+class TestBodies():
+
+    def test_body_init(self):
+        body = Body('body')
+
+        assert body.name is None
+        assert body.mass == Symbol('mass')
```
tarzzz: Does the `mass` assigned automatically on Body Initialization?

pydy/system.py
moorepants: @tarzzz That's a good point. The ability to add forces and torques should be defined first. If we think that gravity warrants a special case and deserves it's own method, then we can add that. Gravity is a case that is use ubiquitously, i.e apply the same force to all bodies/particles in the system.

pydy/tests/test_bodies.py
sahilshekhawat: Yes, every thing is assigned based on the name of the body which provides
us two options when user does not passes a name. One is to make it none and
thus everything else is the default one second is to use assignsymbol
class.  I forgot to change it here but ideally body name should not be none
and mass should be body. Name + '_mass'

sahilshekhawat: You can look here for the correct implementation https://github.com/pydy/pydy/pull/249/files#diff-5727bd04fc961cdc0987b7eba64b51a7R449

pydy/tests/test_system.py
tarzzz: So the new method uses KanesMethod or LangrangesMethod behind the scenes to generate those EoMs?

pydy/system.py
sahilshekhawat: Point noted! I will create a separate class ``Force`` and ``Gravity`` will be a subclass.

pydy/tests/test_system.py
sahilshekhawat: KanesMethod

pydy/tests/test_system.py
```diff
@@ -395,3 +397,67 @@ def test_integrate(self):
...
             sys.generate_ode_function(generator='made-up')
+
+    def test_joint_system_init(self):
+        sys = System(method='joints')
+
+        # test that none of attributes set when system is called with kane
+        # are set.
+        assert hasattr(sys, 'method')
+        assert sys.method == 'joints'
+        assert sys.constants == dict()
```
tarzzz: You should be using `isinstance(sys.constants, dict)` here as well.

sahilshekhawat: noted, will change all such instances. Thanks

pydy/tests/test_system.py
tarzzz: Not sure adding extra classes for this work is a healthy development..
[Stop Writing Classes](http://pyvideo.org/video/880/stop-writing-classes)

sahilshekhawat: I can replace it with a method in System class.

pydy/tests/test_utils.py
```diff
@@ -18,3 +22,61 @@ def test_sympy_equal_to_or_newer_than():
...
+    body = Body('body')
+    body_length = Length(body.name)
+    assert body_length.symbol == Symbol(body.name + "_length")
+
+    body_length_value = Length(body.name, 2)
+    assert body_length_value.value == 2
+
+def test_gravitational_constant():
+
+    gravitational_constant = GravitationalConstant()
```
tarzzz: Another Class? `GravitationalConstant`..?

sahilshekhawat: All of the test_utils is redundant for now. I am trying to push atomic PRs. All of this will be removed there. Thanks

pydy/tests/test_bodies.py
tarzzz: an `OrderedDict` or something should do for a parameter and its constant value pair. If we have require more functionality apart from simply storing the data(params, symbols etc.) we can think of having a separate class, and put the functionality in classmethods.

pydy/system.py
moorepants: Note that we've represented forces and torques as tuples so far. It may only need to be a namedtuple unless you need some methods for Force and Torque.

moorepants:
Ok, let's hold off on this until Sahil breaks it up into atomic PRs.

sahilshekhawat:
I am pretty tired,  can I push them in the morning?

moorepants:
Yes, we are never expecting you to loose sleep here. The only expectation is that you work 40 hrs a week. You get to choose those hours.

pydy/tests/test_joints.py
sahilshekhawat: Yes

pydy/tests/test_system.py
sahilshekhawat: @moorepants Will this ``JointMethod`` be related to ``System`` or ``system`` will just use it like it used Kanes method?

moorepants: I was imagining that the `'JointMethod` would be equivalent in function as `KanesMethod` and `LagrangesMethod`.  Instead of a System class managing all the bodies and joints, the `JointMethod` class would do that. And we could leave our current system class as is. But then the question is where do we include the visualization stuff. I think adding viz decorators only makes sense on the System class, if we are trying to delineate purely symbolic classes from the classes that mix numerics and symbolics.

This is just a brain fart. I'm not sure it is a good idea yet, but I'll think about it this weekend some more. What do you think?

sahilshekhawat: I think its a great idea and we should definitely think about it. I will finish up the unit tests, until then all of us can think about the pro and cons of this implementation.

moorepants: Can you ask for opinion on the mailing list?

sahilshekhawat: yeah, sure.

PR #250, opened 2015-06-28
Base structure of Body class and tests added

sahilshekhawat:
  commit message.
  directory)
  follow deprecation cycles.)

### DIFF
--- a/pydy/bodies.py
+++ b/pydy/bodies.py
@@ -1,6 +1,6 @@
 from sympy import Symbol
 from sympy.physics.mechanics import RigidBody, Particle, ReferenceFrame, outer
-from sympy.physics.vector import Point
+from sympy.physics.vector import Point, Vector
 
 __all__ = ['Body']
 
@@ -95,7 +95,7 @@ class Body(object):
         else:
             self._frame = frame
 
-        if inertia is None:
+        if inertia is None and mass is None:
             inertia = outer(self._frame.x, self._frame.x)  # tensor product
             self._inertia = (inertia, self._masscenter)
         else:
@@ -142,15 +142,15 @@ class Body(object):
         """
         point_vector = self._convert_tuple_to_vector(point_vector)
         force_vector = self._convert_tuple_to_vector(force_vector)
-        point = self._masscenter.locatenew(self.name + '_point' + self._counter,
+        point = self._masscenter.locatenew(self.name + '_point' + str(self._counter),
                                            point_vector)
         self.force_list.append((point, force_vector))
         self._counter += 1
 
     def _convert_tuple_to_vector(self, pos_tuple):
         if len(pos_tuple) == 3:
-            unit_vectors = [self.frame.x, self.frame.y, self.frame.z]
-            vector = 0
+            unit_vectors = [self._frame.x, self._frame.y, self._frame.z]
+            vector = Vector(0)
             for i in range(3):
                 vector += pos_tuple[i] * unit_vectors[i]
             return vector
--- a/pydy/bodies.py
+++ b/pydy/bodies.py
@@ -140,29 +140,18 @@ class Body(object):
         >>> body.add_force((1,0,0), (0,1,0))
 
         """
-        point_vector = self.convert_tuple_into_vector(point_vector)
-        force_vector = self.convert_tuple_into_vector(force_vector)
+        point_vector = self._convert_tuple_to_vector(point_vector)
+        force_vector = self._convert_tuple_to_vector(force_vector)
         point = self._masscenter.locatenew(self.name + '_point' + self._counter, point_vector)
         self.force_list.append((point, force_vector))
         self._counter += 1
 
-    def convert_tuple_into_vector(self, values_tuple):
-        vector = 0
-        vector += vector + values_tuple[0] * self.frame.x
-        vector += vector + values_tuple[1] * self.frame.y
-        vector += vector + values_tuple[2] * self.frame.z
-        return vector
-
-
-class Ground(Body):
-    """
-    A Ground Body
-
-    It has infinite mass and an origin.
-    """
-    def __init__(self):
-        name = 'ground'
-        masscenter = Point('origin')
-        mass = oo
-        # TODO Discuss if Ground needs to be RigidBody or Particle is fine?
-        super(Ground, self).__init__(name, masscenter, mass)
+    def _convert_tuple_to_vector(self, pos_tuple):
+        if len(pos_tuple) == 3:
+            unit_vectors = [self.frame.x, self.frame.y, self.frame.z]
+            vector = 0
+            for i in range(3):
+                vector += pos_tuple[i] * unit_vectors[i]
+            return vector
+        else:
+            raise TypeError('position tuple must be of length 3')
--- a/pydy/bodies.py
+++ b/pydy/bodies.py
@@ -1,11 +1,12 @@
 from sympy import Symbol
-from sympy.physics.mechanics import RigidBody, Particle, ReferenceFrame, outer
+from sympy.physics.mechanics import RigidBody, Particle, ReferenceFrame, \
+    outer, inertia
 from sympy.physics.vector import Point, Vector
 
 __all__ = ['Body']
 
 
-class Body(object):
+class Body(RigidBody, Particle):
     """
     A Body which can be connected by joints.
 
@@ -31,8 +32,8 @@ class Body(object):
         The ReferenceFrame which the rigid body is fixed in.
     mass : Sympifyable (optional)
         The body's mass.
-    inertia : (Dyadic, Point) (optional)
-        The body's inertia about a point; stored in a tuple as shown above.
+    body_inertia : Dyadic (instance of inertia)
+        The body's inertia about center of mass.
 
     Example:
     --------
@@ -46,14 +47,13 @@ class Body(object):
      RigidBody can be passed while creating a Body too.
 
     >>> from sympy import Symbol
-    >>> from sympy.physics.mechanics import ReferenceFrame, Point
+    >>> from sympy.physics.mechanics import ReferenceFrame, Point, inertia
     >>> from pydy.bodies import Body
     >>> mass = Symbol('mass')
     >>> masscenter = Point('masscenter')
     >>> frame = ReferenceFrame('frame')
-    >>> I = outer (A.x, A.x)
-    >>> inertia_tuple = (I, P)
-    >>> body = Body('name_of_body', masscenter, mass, frame, inertia_tuple)
+    >>> body_inertia = inertia(frame, 1, 0, 0)
+    >>> body = Body('name_of_body', masscenter, mass, frame, body_inertia)
 
     3. Creating a Particle. If masscenter and mass are passed, and inertia is
      not then a Particle is created.
@@ -62,58 +62,47 @@ class Body(object):
     >>> from sympy import Point
     >>> from pydy.bodies import Body
     >>> mass = Symbol('mass')
-    >>> masscenter = Symbol('masscenter')
+    >>> masscenter = Point('masscenter')
     >>> body = Body('name_of_body', masscenter, mass)
 
     Similarly, A frame can also be passed while creating a Particle.
 
     """
     def __init__(self, name, masscenter=None, mass=None, frame=None,
-                 inertia=None):
+                 body_inertia=None):
 
-        self.name = name
+        _name = name
         self.parent = None
         self.child = None
         self.force_list = []
-        self.body = None  # TODO assign a better name.
-        self._coordinates = []
-        self._speeds = []
         self._counter = 0
 
         if masscenter is None:
-            self._masscenter = Point(self.name + '_masscenter')
+            self._masscenter = Point(_name + '_masscenter')
         else:
             self._masscenter = masscenter
 
         if mass is None:
-            self._mass = Symbol(self.name + '_mass')
+            _mass = Symbol(_name + '_mass')
         else:
-            self._mass = mass
+            _mass = mass
 
         if frame is None:
-            self._frame = ReferenceFrame(self.name + '_frame')
+            self._frame = ReferenceFrame(_name + '_frame')
         else:
             self._frame = frame
 
-        if inertia is None and mass is None:
-            inertia = outer(self._frame.x, self._frame.x)  # tensor product
-            self._inertia = (inertia, self._masscenter)
+        if body_inertia is None and mass is None:
+            _inertia = (inertia(self._frame, 1, 1, 1), self._masscenter)
         else:
-            self._inertia = inertia
+            _inertia = (body_inertia, self._masscenter)
 
         # If user passes masscenter and mass then a particle is created
         # otherwise a rigidbody. As a result a body may or may not have inertia.
-        if inertia is None and mass is not None:
-            self.body = Particle(self.name, self._masscenter, self._mass)
+        if body_inertia is None and mass is not None:
+            Particle.__init__(self, name, self._masscenter, _mass)
         else:
-            self.body = RigidBody(self.name, self._masscenter, self._frame,
-                                  self._mass, self._inertia)
-
-    def add_coordinate(self, coordinate):
-        self._coordinates.append(coordinate)
-
-    def add_speed(self, speed):
-        self._speeds.append(speed)
+            RigidBody.__init__(self, _name, self._masscenter, self._frame, _mass, _inertia)
 
     def add_force(self, point_vector, force_vector):
         """
@@ -140,19 +129,27 @@ class Body(object):
         >>> body.add_force((1,0,0), (0,1,0))
 
         """
-        point_vector = self._convert_tuple_to_vector(point_vector)
-        force_vector = self._convert_tuple_to_vector(force_vector)
-        point = self._masscenter.locatenew(self.name + '_point' + str(self._counter),
+        if not isinstance(point_vector, tuple):
+            raise TypeError("Point Vector must be a tuple of length 3")
+        else:
+            point_vector = self._convert_tuple_to_vector(point_vector)
+
+        if not isinstance(force_vector, tuple):
+            raise TypeError("Force vector must be a tuple of length 3")
+        else:
+            force_vector = self._convert_tuple_to_vector(force_vector)
+
+        point = self._masscenter.locatenew(self._name + '_point' + str(self._counter),
                                            point_vector)
         self.force_list.append((point, force_vector))
         self._counter += 1
 
     def _convert_tuple_to_vector(self, pos_tuple):
-        if len(pos_tuple) == 3:
+        if len(pos_tuple) != 3:
+            raise TypeError('position tuple must be of length 3')
+        else:
             unit_vectors = [self._frame.x, self._frame.y, self._frame.z]
             vector = Vector(0)
             for i in range(3):
                 vector += pos_tuple[i] * unit_vectors[i]
             return vector
-        else:
-            raise TypeError('position tuple must be of length 3')
--- a/pydy/bodies.py
+++ b/pydy/bodies.py
@@ -58,4 +58,4 @@ class Body(RigidBody):
 class Ground(Body):
     def __init__(self, name, masscenter=None, frame=None, mass=None, inertia=None):
         # TODO change properties of Body.
-        Body.__init__(self, name, masscenter, frame, mass, inertia)
+        super(Ground, self).__init__(*args,**kwargs)
--- a/pydy/bodies.py
+++ b/pydy/bodies.py
@@ -1,6 +1,6 @@
 from sympy import Symbol
 from sympy.physics.mechanics import RigidBody, Particle, ReferenceFrame, \
-    outer, inertia
+    inertia
 from sympy.physics.vector import Point, Vector
 
 __all__ = ['Body']
@@ -97,6 +97,8 @@ class Body(RigidBody, Particle):
         else:
             _inertia = (body_inertia, self._masscenter)
 
+        self._masscenter.set_vel(self._frame, 0)
+
         # If user passes masscenter and mass then a particle is created
         # otherwise a rigidbody. As a result a body may or may not have inertia.
         if body_inertia is None and mass is not None:
@@ -104,7 +106,7 @@ class Body(RigidBody, Particle):
         else:
             RigidBody.__init__(self, _name, self._masscenter, self._frame, _mass, _inertia)
 
-    def add_force(self, point_vector, force_vector):
+    def add_force(self, force_vector, point=None):
         """
         Adds force to the body by adding Force's instance to the force_list.
         force_list is used by system to get the tuples from the instance and
@@ -125,22 +127,20 @@ class Body(RigidBody, Particle):
         To add a force of magnitude 1 in y direction on a point at distance of
         1 in x direction. All the directions are w.r.t body's frame.
 
+        >>> from sympy import Symbol
+        >>> from pydy.bodies import Body
         >>> body = Body('body')
-        >>> body.add_force((1,0,0), (0,1,0))
+        >>> g = Symbol('g')
+        >>> body.add_force((body.get_mass() * g, 0, 0), body.get_masscenter())
 
         """
-        if not isinstance(point_vector, tuple):
-            raise TypeError("Point Vector must be a tuple of length 3")
-        else:
-            point_vector = self._convert_tuple_to_vector(point_vector)
+        if point is None:
+            point = self._masscenter  # masscenter
 
         if not isinstance(force_vector, tuple):
             raise TypeError("Force vector must be a tuple of length 3")
         else:
             force_vector = self._convert_tuple_to_vector(force_vector)
-
-        point = self._masscenter.locatenew(self._name + '_point' + str(self._counter),
-                                           point_vector)
         self.force_list.append((point, force_vector))
         self._counter += 1
 
--- a/pydy/bodies.py
+++ b/pydy/bodies.py
@@ -35,8 +35,9 @@ class Body(RigidBody):
 
     def add_force(self, force):
         """
-        Appends the constant from Force class to the force_list class. Applying exact
-        value of force is handled by the system.
+        Adds force to the body by adding Force's instance to the force_list.
+        force_list is used by system to get the tuples from the instance and
+        create force_list for Kane's method.
 
         Paramters
         ---------
@@ -47,7 +48,8 @@ class Body(RigidBody):
         Example
         -------
         >>> body = Body('body')
-        >>> body.add_force(Force.GRAVITY)
+        >>> force = Force(body.masscenter, body.frame.z, Symbol('g'))
+        >>> body.add_force(force)
 
         """
         # TODO
--- a/pydy/bodies.py
+++ b/pydy/bodies.py
@@ -1,0 +1,59 @@
+from sympy.physics.mechanics import RigidBody
+
+from .force import Force
+
+__all__ = ['Body', 'Ground']
+
+
+class Body(RigidBody):
+    """
+    Paramters
+    ---------
+    name: string
+        defines the name of the body. It is used as the base for defining body specific
+        properties.
+    (Rest are same as RigidBody but optional)
+    masscenter : Point (optional)
+        The point which represents the center of mass of the rigid body.
+    frame : ReferenceFrame (optional)
+        The ReferenceFrame which the rigid body is fixed in.
+    mass : Sympifyable (optional)
+        The body's mass.
+    inertia : (Dyadic, Point) (optional)
+        The body's inertia about a point; stored in a tuple as shown above.
+
+    """
+    def __init__(self, name, masscenter=None, frame=None, mass=None, inertia=None):
+        self.name = name
+        self.parent = None
+        self.child = None
+        self.force_list = []
+
+        # TODO define properties for RigidBody.
+
+        RigidBody.__init__(self, name, masscenter, frame, mass, inertia)
+
+    def add_force(self, force):
+        """
+        Appends the constant from Force class to the force_list class. Applying exact
+        value of force is handled by the system.
+
+        Paramters
+        ---------
+        force: pydy.Force
+            a constant from pydy.Force which will be used by the system to get the exact
+            value of the force.
+
+        Example
+        -------
+        >>> body = Body('body')
+        >>> body.add_force(Force.GRAVITY)
+
+        """
+        # TODO
+
+
+class Ground(Body):
+    def __init__(self, name, masscenter=None, frame=None, mass=None, inertia=None):
+        # TODO change properties of Body.
+        Body.__init__(self, name, masscenter, frame, mass, inertia)
--- a/pydy/bodies.py
+++ b/pydy/bodies.py
@@ -1,8 +1,8 @@
-from sympy import Symbol, oo
+from sympy import Symbol
 from sympy.physics.mechanics import RigidBody, Particle, ReferenceFrame, outer
 from sympy.physics.vector import Point
 
-__all__ = ['Body', 'Ground']
+__all__ = ['Body']
 
 
 class Body(object):
@@ -142,7 +142,8 @@ class Body(object):
         """
         point_vector = self._convert_tuple_to_vector(point_vector)
         force_vector = self._convert_tuple_to_vector(force_vector)
-        point = self._masscenter.locatenew(self.name + '_point' + self._counter, point_vector)
+        point = self._masscenter.locatenew(self.name + '_point' + self._counter,
+                                           point_vector)
         self.force_list.append((point, force_vector))
         self._counter += 1
 
--- a/pydy/bodies.py
+++ b/pydy/bodies.py
@@ -1,19 +1,30 @@
-from sympy.physics.mechanics import RigidBody, Particle
-
-from .force import Force
+from sympy import Symbol, oo
+from sympy.physics.mechanics import RigidBody, Particle, ReferenceFrame, outer
+from sympy.physics.vector import Point
 
 __all__ = ['Body', 'Ground']
 
 
 class Body(object):
     """
+    A Body which can be connected by joints.
+
+    It can create a Particle or RigidBody depending upon the arguments passed.
+    If nothing is passed, then by default, a RigidBody is created. User can
+    also pass masscenter, mass, frame and inertia to change the arguments
+    passed to RigidBody.
+    To create a Particle only mass and masscenter needs to be passed. User can
+    also pass a frame which will be attached to the Particle.
+
+    In short, it creates a Rigidbody if inertia is passed or nothing is passed.
+    and a Particle if inertia is not passed and things like masscenter and mass
+    are.
 
     Parameters
     ---------
     name: string
-        defines the name of the body. It is used as the base for defining body specific
-        properties.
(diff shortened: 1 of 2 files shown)


pydy/bodies.py
```diff
@@ -0,0 +1,59 @@
+from sympy.physics.mechanics import RigidBody
+
+from .force import Force
+
+__all__ = ['Body', 'Ground']
+
+
+class Body(RigidBody):
```
moorepants: Shouldn't a Body be able to be a derived from both a Particle or a RigidBody?

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,50 @@
...
+        name = self.body.name
+        assert self.body.masscenter == Point(name + '_MassCenter')
+        assert self.body.frame == ReferenceFrame(name + '_ReferenceFrame')
+        assert self.body.mass == Symbol(name + '_Mass')
+
+    def test_body_add_force(self):
+        assert isinstance(self.body.force_list, list())
+        assert len(self.body.force_list) == 0
+
+        self.body.add_force(Force.GRAVITY)
```
moorepants: What is "Force.GRAVITY"?

moorepants: Is it possible to have a method `add_load` that can add either a torque or a force?

sahilshekhawat: Its a constant in Force class, Instead of create classes we will implement different forces as methods. e.g. Force.gravity() but it requires details of the body, thus I needed a constant so that I can add the exact force in system class.

moorepants: How will you deal with equal and opposite loads? For, example, if you add a spring force between two points on a body, then there are equal and opposite forces.

moorepants: A force is a vector applied to a point. That's the big generalization. I think it would be fine to create `class Load(object)`, `class Force(Load)`, and `class Torque(Load)` (not sure if Load is necessary) classes. But maybe these classes should somehow be smart enough to apply equal and opposite forces. You could initalize them like `Force(vec, point1, opp_point=point2)` or something.

sahilshekhawat: As system is handling forces/torques, I think it is possible to add a load.  ie..create a method add_load instead of add_force

pydy/bodies.py
sahilshekhawat: I intend to create a separate subclass of Particle to incorporate things required to build joints system.

pydy/tests/test_bodies.py
moorepants: The unit test should show where `Force.GRAVITY` is derived from. For example,

```python
g = sm.symbols('g')
force = Force(g * ref_frame.y, *all_points)
force.GRAVITY
```

pydy/bodies.py
moorepants: Why?

pydy/tests/test_bodies.py
sahilshekhawat: Let me explain what I have done here, I apologize if all of this is confusing. I think I should have raised a PR for Force too.
In our yesterday's meeting tarun and I decided that instead of using Force's constructor to pass body and other paramters lets do body.add_force() which have problem that I can't access body's details like mass. I have asked for the help on gitter, until tarun sees that as a quick fix I did this implementation.
Force have methods like gravity and constants like GRAVITY which can be anything e.g. a string 'gravity' like an enum and there is no constructor of Force. There is one method and one constant for each individual force. System use Force.GRAVITY to add the force to the body and body just holds the constant.

moorepants: But forces are not constants. In the general sense, a force is a time varying 3D vector applied to a point.

And in the unit test you have `Force.GRAVITY` but you never define `Force` above anywhere. It isn't even imported.

sahilshekhawat: I have imported Force class. Please refer here https://github.com/pydy/pydy/pull/250#discussion-diff-33424937R11

moorepants:
Note that the concept of force in sympy.physics.mechanics is simple a tuple of a Vector and a Point, e.g.: `tuple(Vector, Point)`. That encompasses the essence of a force.

pydy/tests/test_bodies.py
sahilshekhawat: Please let me raise a PR for the Force class to clear the doubts.

moorepants: > It isn't even imported.

Sorry I see that it is imported above.

moorepants:
> It isn't even imported.

Sorry I see that it is imported above.

moorepants:
> Please let me raise a PR for the Force class to clear the doubts.

That's fine, but for future PRs, you should write a paragraph in the very first comment window explaining what this PR is all about, what things are assumed, etc. Give us some explanation of what you are presenting. Right now I just see this and my comments may not be fruitful because I don't understand the bigger picture.

sahilshekhawat:
For a quick review. Below is an extract from the implementation of Force class.
```
class Force():

    GRAVITY = 'gravity'  # can be anything, its just constant

    def gravity(self, body, direction_vector):
        g = Symbol('g')
        return body.mass * g * direction_vector
```

moorepants:
I don't think a Force class should have anything about gravity. The gravitational force is one of an infinite number of and types of forces that can be applied. Gravity is only special because it is a force that is often common to every single body in the system and it is of constant magnitude and direction.  A class that represents a Force should be general.

sahilshekhawat:
I would like to point out that this implementation was discussed in yesterdays meeting with Tarun, I have come up with this constant thing to resolve an issue. 
For now, can I remove this add force method. I will add this with the implemetation of Force class?

sahilshekhawat:
And to address infinite types of forces, we only need to support some types of forces only and all the other forces will be handled with ``Force.custom()`` method.

moorepants:
I'll wait until you present the unit tests for the Force class before saying anything more.

pydy/bodies.py
```diff
@@ -0,0 +1,61 @@
...
+        >>> body.add_force(force)
+
+        """
+        # TODO
+
+
+class Ground(Body):
+    def __init__(self, name, masscenter=None, frame=None, mass=None, inertia=None):
+        # TODO change properties of Body.
+        Body.__init__(self, name, masscenter, frame, mass, inertia)
```
tarzzz: You should use:
    
     super(Ground, self).__init__(*args,**kwargs)

to initialize sub-classes..

sahilshekhawat: Okay, changing.

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,51 @@
...
+        assert isinstance(self.body.mass, Symbol)
+        assert len(self.body.inertia) == 2
+        assert isinstance(self.body.inertia[0], Dyadic)
+        assert isinstance(self.body.inertia[1], Point)
+
+    def test_body_properties_value(self):
+        name = self.body.name
+        assert self.body.masscenter == Point(name + '_MassCenter')
+        assert self.body.frame == ReferenceFrame(name + '_ReferenceFrame')
+        assert self.body.mass == Symbol(name + '_Mass')
```
tarzzz: I am slightly skeptical on this naming convention. @moorepants Should we be using Camelcases for naming prefixes?
Although a very minute thing, but would save sahil some rework if we decide on it before he implements it.

pydy/bodies.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    def __init__(self, name, masscenter=None, frame=None, mass=None, inertia=None):
+        self.name = name
+        self.parent = None
+        self.child = None
+        self.force_list = []
+
+        # TODO define properties for RigidBody.
+
+        RigidBody.__init__(self, name, masscenter, frame, mass, inertia)
```
tarzzz: Should be:

    super(Body, self).__init__(*args,**kwargs)

pydy/tests/test_bodies.py
moorepants: We have "names" on almost all of our objects but besides printing unit vectors and some of the objects to the screen I'm not even sure what all these names are for. I could imagine a checkbox in the visualizer that could show the names of different objects.

This is just a string, it can be anything we want, but I would recommend being consistent with whatever is chosen.

sahilshekhawat: Names will be important in the GUI, as users will have to change numeric values.

moorepants: For the Symbol, it is more typical to use mathematical names. Kane has a pretty nice and orderly convention for naming reference frames, unit vectors, etc. in his book.

moorepants: If the point of all these names is to be displayed in the GUI, then maybe more human readable names would be preferred, i.e. no underscores, no camel case. `"Cart", "Cart Mass Center", "Cart Reference Frame"`.

moorepants: This also begs the question as to whether each object should have a mathematical name and a human readable name. In SymPy things like this are handled by printing different representations of the objects.

pydy/bodies.py
```diff
@@ -0,0 +1,73 @@
...
+    (Rest are same as RigidBody but optional)
+    masscenter : Point (optional)
+        The point which represents the center of mass of the rigid body.
+    frame : ReferenceFrame (optional)
+        The ReferenceFrame which the rigid body is fixed in.
+    mass : Sympifyable (optional)
+        The body's mass.
+    inertia : (Dyadic, Point) (optional)
+        The body's inertia about a point; stored in a tuple as shown above.
+
```
moorepants: Seems to me that a Body should accept a RigidBody or a Particle + a ReferenceFrame. Why did you not decide to do that?

pydy/bodies.py
```diff
@@ -0,0 +1,73 @@
...
+        # TODO define properties for RigidBody.
+
+        # is user passes masscenter and mass then a particle is created otherwise a rigidbody.
+        # as a result a body may or may not have inertia.
+        if inertia is None and mass is not None:
+                self.body = Particle(self.name, self.masscenter, self.mass)
+        else:
+                self.body = RigidBody(self.name, self.masscenter, self.frame, self.mas
+
+    def add_force(self, force):
```
moorepants: Unless you have a facility for adding arbitrary points to a body, then I don't think we should be adding forces to a body.

pydy/bodies.py
sahilshekhawat: I did. It creates an instance of particle if only mass and point are provided  and RigidBody otherwise.

sahilshekhawat: Frame is added to a Particle after that.

pydy/bodies.py
```diff
@@ -0,0 +1,73 @@
...
+        Paramters
+        ---------
+        force: (point, vector) or vector
+            Adds the vector force to the point. point must have to be in the body.
+            If body is RigidBody, pass (point, vector) and if body is Particle then pass vector,
+            force is always applied to the masscenter for particle.
+        """
+        # TODO
+
+    def add_coordinate(self, coordinate):
```
moorepants: Let's avoid methods that start with 'add' or 'set' or 'subtract' or 'get. Just use the `@property` decorator.

pydy/bodies.py
moorepants: What I mean is why isn't the api either:

```python
body1 = Body(my_rigidbody_instance)
```
or 
python```
body2 = Body(particle_instance, ref_frame_instance)
```

Comment (moorepants) [inline pydy/bodies.py]: Those classes already store all the information you need. So why make it redundant?

File: pydy/bodies.py (hunk 7 above)
Comment (sahilshekhawat) [inline pydy/bodies.py]: It defines a point relative to body's masscenter and add force to it. The passed value is a vector similar to what we are doing in joints. (passing as a 3 tuple)

File: pydy/bodies.py (hunk 8 above)
Comment (sahilshekhawat) [inline pydy/bodies.py]: Acknowledged.

File: pydy/tests/test_bodies.py [hunk 9]
```diff
@@ -0,0 +1,80 @@
...
+
+class TestBody():
+    def setup(self):
+        self.body = Body('Body')
+
+    def test_body_init(self):
+        # by default a rigidbody is created.
+        assert self.body.name == 'Body'
+        assert self.body.parent is None
+        assert self.body.child is None
```
moorepants: Why do parent child relationships need to be stored on body? I thought this was stored on Joint. I don't think it is necessary to store it both on Joint and Body.

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,80 @@
...
+        assert isinstance(self.body.masscenter, Point)
+        assert isinstance(self.body.frame,    ReferenceFrame)
+        assert isinstance(self.body.mass, Symbol)
+        assert len(self.body.inertia) == 2
```
moorepants: ? Shouldn't this be an inertia dyadic?

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,80 @@
...
+        assert self.body.frame == ReferenceFrame(name + ' ReferenceFrame')
+        assert self.body.mass == Symbol(name + ' Mass')
+
+    def test_body_particle(self):
+        point = Point('point')
+        mass = Symbol('mass')
+        self.particle_body = Body(masscenter=point, mass=mass)
+        assert isinstance(self.particle_body.body, Particle)
+        assert self.particle_body.masscenter == point
+        assert self.particle_body.mass == mass
```
moorepants: How is the reference frame generated?

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def test_body_particle_add_force(self):
```
moorepants: It may be useful for the user to add a force easily to the particle point or the mass center of a rigidbody, but in general you need to be able to apply forces to arbitrary points.

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,80 @@
...
+        point = self.body.masscenter.locatenew('force_point', Symbol('l')*self.body.frame.x)
+        force_tuple = (point, Symbol('Fa')*self.body.frame.z)
+        self.body.add_force(force_tuple)
+        assert len(self.body.force_list) == 1
+        assert self.body.force_list[0] == [force_tuple]
+
+    def test_body_coordinates(self):
+        assert self.body.coordinates == []
+        q1 = dynamicsymbols('q1')
+        self.body.add_coordinate(q1)
```
moorepants: Why not, `body.coordinates.append(q1)`?

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,80 @@
...
+        assert q1 in self.body.coordinates
+
+    def test_body_speeds(self):
+        assert self.body.speeds == []
+        u1 = dynamicsymbols('u1')
+        self.body.add_speed(u1)
```
moorepants: What is the purpose of `add_speed` and `add_coordinate`? Coordinates are not really associated with bodies. Coordinats are associates with joints. So I'm not sure what the purpose is here.

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,80 @@
...
+        assert u1 in self.body.speeds
+
+
+class TestGround():
```
moorepants: How is Ground different than Body? If there is no difference, a sub class is not needed.

pydy/bodies.py
sahilshekhawat: Then the user have to define masscenter, frame etc. as well since they are required by rigidbody.

pydy/tests/test_bodies.py
sahilshekhawat: It is not stored in joints. Joints sets these attributes.

pydy/tests/test_bodies.py
sahilshekhawat: Every body have its own reference frame, thus, I just create a new reference frame during the instantiation of Body.

pydy/bodies.py
moorepants: Ok, I see. So is this force point relative to the body's reference frame or any general vector? Does the point have to be fixed to the body?

If this is an arbitrary force and point I don't see why it should be associated with a particular body.

pydy/bodies.py
moorepants: Sometimes they are needed, but it is likely you can simply use the methods on the attribute to do what you need: `body.coordinates.append()`.

pydy/bodies.py
moorepants: The question is does a user create a RigidBody first and pass to Body or just create a Body?

pydy/tests/test_bodies.py
moorepants: Ok, I misunderstood. That may be fine then.

pydy/tests/test_bodies.py
moorepants: So shouldn't there be a test for that here?

pydy/tests/test_bodies.py
sahilshekhawat: User cannot have access to coordinates, this might cause a lot of problems. so, I want to keep internal working private.

moorepants: Ok, then make it a private attribute and private method.

pydy/bodies.py
sahilshekhawat: Just create a body. Please see my latest changes regarding this. Thank you.

pydy/bodies.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+    A Ground Body
+
+    It has infinite mass and an origin.
+    """
+    def __init__(self):
+        name = 'ground'
+        masscenter = Point('origin')
+        mass = oo
+        # TODO Discuss if Ground needs to be RigidBody or Particle is fine?
```
sahilshekhawat: Right now, Ground is a Particle with infinite mass, should this be RigidBody instead of Particle is just fine?

moorepants: I'm not sure that we even need a special ground body in the first place. If you do need it, this infinite mass will not come into any computations, so I don't think it is needed at all.

pydy/bodies.py
```diff
@@ -0,0 +1,168 @@
...
+        else:
+            self._masscenter = masscenter
+
+        if mass is None:
+            self._mass = Symbol(self.name + '_mass')
+        else:
+            self._mass = mass
+
+        if frame is None:
+            self._frame = ReferenceFrame(self.name + '_frame')
```
moorepants: Instead of creating a ReferenceFrame and Point, etc, shouldn't a RigidBody or Particle be created in the class. That way you will  remove redundancy and can get all of the other methods and attributes on RigidBody and Particle for free.

I'm still not sold on this body class as is.

I'm wondering if it should somehow be a super class of rigidbody and particle...

pydy/bodies.py
```diff
@@ -0,0 +1,168 @@
...
+        else:
+            self._inertia = inertia
+
+        # If user passes masscenter and mass then a particle is created
+        # otherwise a rigidbody. As a result a body may or may not have inertia.
+        if inertia is None and mass is not None:
+            self.body = Particle(self.name, self._masscenter, self._mass)
+        else:
+            self.body = RigidBody(self.name, self._masscenter, self._frame,
+                                  self._mass, self._inertia)
```
moorepants: Ah! It does do that. Ok, that's good.

moorepants: I don't see any tests that test this.

sahilshekhawat: I am really bad with TDD, will try to improve on that.

pydy/bodies.py
sahilshekhawat: RigidBody and Particle are created in body. Do I explain it more?

pydy/bodies.py
sahilshekhawat: Yes, Ground class is expendable.

pydy/bodies.py
sahilshekhawat: I have updated the tests.

sahilshekhawat:
@pydy/pydy-developers This PR needs a review. Thanks

sahilshekhawat:
Also, I am not sure why tests for last commit are failing on travis.

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,112 @@
...
+from sympy import Symbol
+from sympy.physics.vector import Point, Dyadic, ReferenceFrame
+from sympy.physics.mechanics import Particle, RigidBody,\
+    dynamicsymbols, outer
+
+# local
+from ..bodies import Body
+
+
+def test_points(point1, point2):
```
moorepants: How do these tests run automatically if they need arguments?

sahilshekhawat: These are functions used to test two points. I have used these several times in TestBody class. They are not standalone tests, just utility functions to remove redundancy.

moorepants: Nose will try to run them if the test name starts with "test".

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,112 @@
...
+    assert point1.name == point2.name
+    assert point1._pos_dict == point2._pos_dict
+    assert point1._vel_dict == point2._vel_dict
+    assert point1._acc_dict == point2._acc_dict
```
moorepants: Is it not possible to encode this in Point's equality check?

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,112 @@
...
+
+class TestBody():
+    def test_default(self):
+        body = Body('body')
+        assert body.name == 'body'
+        assert body.parent is None
+        assert body.child is None
+        assert body.force_list == []
+        assert body._coordinates == []
+        assert body._speeds == []
```
moorepants: It still isn't clear to me how coordinates and speeds can be associated with a body. In the current design it seems that coordinates and speeds should stick with the joints. Also, should you be storing joints on bodies?

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,112 @@
...
+        test_points(body._masscenter, Point('body_masscenter'))
+        assert body._mass == Symbol('body_mass')
```
moorepants: Why is the mass private?

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,112 @@
...
+        test_referenceframes(body._frame, ReferenceFrame('body_frame'))
+        assert body._inertia == (outer(body._frame.x, body._frame.x),
```
moorepants: Why is this the default inertia, should it be `inertia(body_frame, ixx, iyy, izz, ixz, ixy, iyz)`?

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,112 @@
...
+                                 body._masscenter)
+        assert isinstance(body.body, RigidBody)
+
+        rigidbody = body.body
```
moorepants: This is odd. Why would a body have a body? Maybe a different attribute name is needed. Also, why make this public? I'd rather see all of the attrs and methods of rigidbody/particle be first class attrs/methods on Body.

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,112 @@
...
+        test_points(rigidbody.masscenter, body._masscenter)
```
moorepants: Once again, why make these attrs private?

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,112 @@
...
+        assert rigidbody.inertia == body._inertia
+
+    def test_custom_rigid_body(self):
+        masscenter = Point('rigidbody_masscenter')
+        mass = Symbol('rigidbody_mass')
+        frame = ReferenceFrame('rigidbody_frame')
+        inertia = outer(frame.x, frame.x)
+        inertia_tuple = (inertia, masscenter)
+        self.rigidbody_body = Body('rigidbody_body', masscenter, mass, frame,
+                                   inertia_tuple)
```
moorepants: If a body's inertia is always defined with respect to it's mass center and reference frame, we shouldn't allow the user to pass in a generic inertia dyadic and point. You should simple accept the same arguments as the `inertia` function, minus the reference frame.

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,112 @@
...
+        assert self.rigidbody_body._mass == mass
+        test_referenceframes(self.rigidbody_body._frame, frame)
+        assert self.rigidbody_body._inertia == inertia_tuple
+        assert isinstance(self.rigidbody_body.body, RigidBody)
+
+    def test_particle_body(self):
+        masscenter = Point('particle_masscenter')
+        mass = Symbol('particle_mass')
+        frame = ReferenceFrame('particle_frame')
+        self.particle_body = Body('particle_body', masscenter, mass, frame)
```
moorepants: Why do you set these to `self`? Do you use the same thing in other tests?

moorepants: If so, shouldn't you just put that in `setup`?

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,112 @@
...
+        assert self.particle_body._mass == mass
+        test_referenceframes(self.particle_body._frame, frame)
+        assert self.particle_body._inertia is None
+        assert isinstance(self.particle_body.body, Particle)
+
+    def test_add_coordinates(self):
+        for body in [self.rigidbody_body, self.particle_body]:
+            assert body._coordinates == []
+            q1 = dynamicsymbols('q1')
+            body.add_coordinate(q1)
```
moorepants: This method needs more explanation. Why do you need to store coordinates on a body?

Also, I'm not found of this `add_` method. If `_coordinates` is a list, why not just use the append method?

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,112 @@
...
+            assert q1 in body._coordinates
+
+    def test_add_speeds(self):
```
moorepants: Same comments I apply to coordinates apply here.

pydy/tests/test_bodies.py
moorepants: Also, note that speeds are explicitly defined in KanesMethod, but in LagrangesMethod and others the "speeds" are simply the derivatives of the coordinates. So maybe speeds should be automatically generated if the are not specified.

moorepants:
I didn't see any test for the parent/child functionality. I'm not sure how that works.

You someone need the attrs/methods of Particle/RigidBody to be first class citizens in the Body class so users don't have to dig for them. It almost seems like Body should be a super class for RigidBody and Particle and the user should then create either a Particle or RigidBody, leaving Body as an abstract base class.

This class is more like a rigidbody/particle generator than a pure "Body" object as it stands.

It wasn't apparent to me, but I assum you effectively associate a reference frame with a particle. That is necessary, no?

moorepants:
@pydy/pydy-developers Sahil would like some comments on his Body class design.

pydy/tests/test_bodies.py
sahilshekhawat: Okay, I will change the names of these funct Thanks

pydy/tests/test_bodies.py
sahilshekhawat: Yes, it is. I was not sure if we want to equate points with same attributes as equal or not. So, I just included it here. Should we add this to Point instead?

pydy/tests/test_bodies.py
sahilshekhawat: Right now, all the coodinates and speeds are associated to bodies. This way we don't have to store joints. I can see now why they must be attached to the Joints. I will change this but for that we need to store joints as well. Not sure if storing on bodies is a good idea.

pydy/tests/test_bodies.py
sahilshekhawat: because all the attributes of RigidBody are accessible and I don't want to confuse users with two masses which are identical. So, if a user want to get mass, he can do ``body.body.mass``. Attribute's name being 'body' needs to be changed though.

pydy/tests/test_bodies.py
sahilshekhawat: I tried reading various texts about dyadics and inertia but was not able to determine default inertia. Thanks, I will change it.

pydy/tests/test_bodies.py
sahilshekhawat: Yes, I should. Changing these things  to setup.

pydy/tests/test_bodies.py
sahilshekhawat: Oh! I didn't see that. Yeah, it will be good. thanks

pydy/tests/test_bodies.py
sahilshekhawat: Yes, I have added a TODO in Body class to change the attribute's name.

sahilshekhawat: Also, there are many attributes and functions in RigidBody which are not there in Body like kinetic_energy . Thus, I don't want to make this attribute private. Instead I made other attributes of Body private so that user just have to know about RigidBody.

pydy/tests/test_bodies.py
moorepants: It seems reasonable to add an `__eq__` method to Point for these checks.

But this also begs the question of whether this is really what you need to check in the tests. If you are creating new points inside the body class then testing for this equality seems necessary. But maybe all you have to do is check if two Points produce the same result from `pos_from` or `vel` call.

pydy/tests/test_bodies.py
moorepants: Joints definitely need to contain references to Bodies as currently designed (you pass in two bodies to a joint, right?). If Joints know about all the bodies and Joints are passed into JointsMethod, then all the information is available. So there is no need to have Joints or coordinates attached to Bodies. But you could also imagine a model where you add Joints to bodies and then pass in Bodies to JointsMethod. The Joints also can encode the parent child relationship so that may not be necessary to add to Body either.

pydy/tests/test_bodies.py
moorepants: That's more confusing from my perspective. You should be able to do this:

```python
b = Body('b')
b.mass
b.inertia
b.potential_energy()
etc
```

I think the `body` attribute should be private `_body`.

pydy/tests/test_bodies.py
moorepants: A default inertia is a 3 x 3 symmetric tensor/matrix...6 variables. A Dyadic is a form of the tensor which encodes the reference frame. A Dyadic is to a 3 x 3 matrix as a Vector is to a 3 x 1 column matrix (in sympy)

pydy/tests/test_bodies.py
moorepants: But this then becomes confusing. A Body should have a mass, it should have an inertia, etc. Either these should be promoted to the Body class level or we should think about building a sub/super class structure so the right things are inherited.

moorepants:
New though: if you store parent/child info on Joints, maybe Body is not necessary at all. Simply pass in Particles and RigidBodies to Joints. We could also potentially ignore Particle. A particle could be represented by as a RigidBody with the inertia dyadic set to the zero dyadic.

pydy/bodies.py
```diff
@@ -0,0 +1,158 @@
...
+        Example
+        --------
+        To add a force of magnitude 1 in y direction on a point at distance of
+        1 in x direction. All the directions are w.r.t body's frame.
+
+        >>> body = Body('body')
+        >>> body.add_force((1,0,0), (0,1,0))
+
+        """
+        point_vector = self._convert_tuple_to_vector(point_vector)
```
tarzzz: You should put in some checks here, to see if users are supplying the tuple. 
The purpose of this check is to provide a more informative error/tracebacks. 
something like 'This argument should be a tuple..'

pydy/bodies.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def _convert_tuple_to_vector(self, pos_tuple):
+        if len(pos_tuple) == 3:
+            unit_vectors = [self._frame.x, self._frame.y, self._frame.z]
+            vector = Vector(0)
+            for i in range(3):
+                vector += pos_tuple[i] * unit_vectors[i]
+            return vector
+        else:
+            raise TypeError('position tuple must be of length 3')
```
tarzzz: Its better to place checks at the starting of the calling method, rather than at the very end, as it reduces overhead.

sahilshekhawat: Agreed! Changing.

sahilshekhawat:
@moorepants Thats a great idea. I can see how that can be done. Joints will build up parent, child relationship. Even though we then have to store all Joints, it will be simple and elegant. There will be just Joints and JointsMethod. Also, I think that coordinates and speeds are better off within Joints than in Body. I will push the samples by tonight. Thanks

moorepants:
I haven't thought out the idea completely so you should investigate first. If you don't do that then Bodies need to store joints which store coordinates. A body can have many joints or a joint can have two bodies. Pick one idea or the other.

pydy/tests/test_bodies.py
sahilshekhawat: Hmm..that makes sense, I don't even need to check the entire point. Thanks, this one will help a lot.

pydy/tests/test_bodies.py
sahilshekhawat: I have updated the API, not instead of passing Bodies we need to pass Joints. As you mentioned joints will contain all the informations when passed in JointsMethod.

sahilshekhawat: Also, adding joints makes more sense than adding bodies to system. It will eliminate some ambiguous conditions like passing body without any joints and thus, no relations with other bodies. How that will be represented.

pydy/tests/test_bodies.py
sahilshekhawat: For that I need to use dynamic inheritance. I will take Particle or RigidBody as base class depending upon the arguments passed. I am not sure if its a correct design pattern but it will surely fix the problem. Other than this, I don't see how we can have all attributes of Particle or RigidBody depending upon arguments.

sahilshekhawat: Oh! I am sorry. This can be handled by just multiple inheritances.

sahilshekhawat:
Body is now a subclass of Particle and RigidBody, thus, all the attributes of RigidBody and Particle are there. Coordinates and speed will be added to Joints and Joints will be passed to Joints Method instead of body.

pydy/tests/test_bodies.py
```diff
@@ -0,0 +1,90 @@
...
+        l = Symbol('l')
+        Fa = Symbol('Fa')
+        point = self.rigid_body.get_masscenter().locatenew(
+            'rigidbody_body_point0',
+            l * self.rigid_body.get_frame().x)
+        self.rigid_body.add_force((l, 0, 0), (0, 0, Fa))
+        assert len(self.rigid_body.force_list) == 1
+        force_vector = Fa * self.rigid_body.get_frame().z
+        check_points(self.rigid_body.force_list[0][0], point)
+        assert self.rigid_body.force_list[0][1] == force_vector
```
moorepants: Now that you have the multiple inheritance you'll need some tests to ensure that all the methods and attributes work. You could test to see if body will give the same result as the particle and rigidbody tests.

sahilshekhawat: Acknowledged.

moorepants:
This has been submitted to SymPy. Closing.

PR #251, opened 2015-06-29
Created basic structure of force classes

sahilshekhawat:
Works on the basic principle of adding a force with a magnitude and direction to a point. Have a method to return the tuple, magnitude * direction and point. Direction is always taken as a unit vector.

**Parameters**
1. point: sympy.physics.vector.Point - Point of application of force.
2. direction: sympy.physics.vector.Vector - Direction of application of force.
3. magnitude: (optional) - defines the magnitude of force.

**Example**
```
>>> body = Body('body')
>>> force = Force(body.masscenter, body.frame.z, Symbol('g'))
>>> body.add_force(force)
```

  commit message.
  directory)
  follow deprecation cycles.)

### DIFF
--- a/pydy/forces.py
+++ b/pydy/forces.py
@@ -1,0 +1,35 @@
+from sympy import Symbol
+
+__all__ = ['Force', 'GravitationalForce']
+
+
+class Force(object):
+    """
+    It provides an concept of force, although actual force is added by system.
+
+    Parameters
+    ---------
+    point: sympy.physics.vector.Point
+        Point of application of force.
+    direction: sympy.physics.vector.Vector
+        Direction of application of force.
+    magnitude: (optional)
+        defines the magnitude of force.
+
+    Example
+    -------
+    >>> body = Body('body')
+    >>> force = Force(body.masscenter, body.frame.z, Symbol('g'))
+    >>> body.add_force(force)
+
+    """
+    def __init__(self, point, direction, magnitude=None):
+        self.point = point
+        self.direction = direction.normalize()
+        if magnitude is None:
+            self.magnitude = Symbol("ForceMagnitude_" + point.name)
+        else:
+            self.magnitude = magnitude
+
+    def get_force_tuple(self):
+        return (self.magnitude * self.direction, self.point)
--- a/pydy/tests/test_forces.py
+++ b/pydy/tests/test_forces.py
@@ -1,0 +1,21 @@
+from sympy import Symbol
+from sympy.physics.vector import Point
+from sympy.physics.mechanics import ReferenceFrame
+
+from ..forces import Force
+
+class TestForce():
+    def setup(self):
+        self.point = Point('point')
+        self.reference_frame = ReferenceFrame('referenceframe')
+        self.point.set_vel(self.reference_frame, 0)
+        self.force = Force(self.point, self.reference_frame.z, Symbol('g'))
+
+    def test_force_init(self):
+        assert self.force.point == self.point
+        assert self.force.direction == self.direction.normalize()
+        assert self.force.magnitude == self.magnitude
+
+    def test_force_get_force_tuple(self):
+        force_tuple = self.force.get_force_tuple()
+        assert force_tuple = (self.direction.normalize() * self.magnitude, self.point)


pydy/forces.py
```diff
@@ -0,0 +1,35 @@
+from sympy import Symbol
+
+__all__ = ['Force', 'GravitationalForce']
+
+
+class Force(object):
+    """
+    It provides an concept of force, although actual force is added by system.
```
moorepants: This sentence is not very clear. I'm not sure what it means and is it even necessary to mention system here?

pydy/forces.py
```diff
@@ -0,0 +1,35 @@
...
+
+    Parameters
+    ---------
+    point: sympy.physics.vector.Point
+        Point of application of force.
+    direction: sympy.physics.vector.Vector
+        Direction of application of force.
+    magnitude: (optional)
```
moorepants: A Vector already has a magnitude. What is the reason to have this redundant information?

pydy/tests/test_forces.py
```diff
@@ -0,0 +1,21 @@
...
+    def setup(self):
+        self.point = Point('point')
+        self.reference_frame = ReferenceFrame('referenceframe')
+        self.point.set_vel(self.reference_frame, 0)
+        self.force = Force(self.point, self.reference_frame.z, Symbol('g'))
+
+    def test_force_init(self):
+        assert self.force.point == self.point
+        assert self.force.direction == self.direction.normalize()
+        assert self.force.magnitude == self.magnitude
```
moorepants: Vector.magnitude() already can return this.

pydy/tests/test_forces.py
```diff
@@ -0,0 +1,21 @@
...
+from sympy.physics.vector import Point
+from sympy.physics.mechanics import ReferenceFrame
+
+from ..forces import Force
+
+class TestForce():
+    def setup(self):
+        self.point = Point('point')
+        self.reference_frame = ReferenceFrame('referenceframe')
+        self.point.set_vel(self.reference_frame, 0)
```
moorepants: Why is this necessary to instantiate a force?

moorepants:
Right now this class doesn't add anything that a a simple tuple of a Vector and a Point encompass. At the most you could create a namedtuple to do everything this class does. Unless you plan to add other useful attributes and methods to a Force, I don't see any reason for the class.

Secondly, this is likely something that belongs in SymPy. I could image have a Force and Torque class in sympy.physics.mechanics that can be used with all of the method there, instead of passing around a tuple.

Bare minimum, you can do this:

```python
# create a constant force acting on a moving point
Force = namedtuple('Force', ['vector', 'point'])
N = ReferenceFrame('N')
a, b, c = symbols('a, b, c')
q1, q2, q3 = dynamicsymbols('q1, q2, q3')
O = Point('O')
O.set_vel(N, 0)
P = O.locate_new('P', q1 * N.x + q2 * N.y + q3 * N.z) 
force = Force(a * N.x + b * N.y + c * N.z, P)
```

Some concepts that you could possibly use to make having a Force and Torque class be worth while:

- Forces and Torques always have to have equal and opposite reactions between bodies. For example, if you specify a spring force that represents a spring that is connected to two points then you must apply an equal and opposite force to each point. Right now, in our API you have to specify each of those forces separately. Maybe be a Force/Torque class could allow you to generate both forces from one method call.
- It could be useful to name a Force/Torque.
- It would be useful to visualize forces and torques. In general we'd like to visualize vectors, but forces would be visualized attached to a point but changing in magnitude and direction.
- It would be nice nice if Force /Torque had methods that computed their contribution to a partial velocity. Every force/torque can be dotted with the partial velocities to get the generalized active force.
- It could be useful to know if forces/torques do work or not.
- Just like you will have Joints, you could objects that represent certain types of ForceGenerators. A spring is a force generator. So a Spring class may take two points and a spring function. A LinearSpring would take two points and a spring constant (symbol or number). Both of these should then return the two Force objects. You can also have standard Dampers or Gravity (adds constant force to all bodies/particles) or Fluid or Air (adds drag), etc. All of these are either actual objects that generate specific types of forces/torques. But first you need to have a well defined Force/Torque concept.

moorepants:
Also note that adding a Force to a Body is not a complete well specified thing. The only Point that is guaranteed to be associated with a Body is it's mass center. It is true that if you want to apply a force to a point elsewhere on the body that you can specify and equivalent force at the mass center and a couple. So if you want `add_force` to be useful it would need to accept a force and a tuple. But that puts more burden than is necessary on the user. The user would prefer to define other points explicitly to apply a force to. There are some other special points though. There are an infinite number of points that are fixed in the body relative to the mass center. These points could warrent a method like:

```python
body = Body(...)
body.add_body_fixed_point((1, 2, 3))
```

This would add a point relative to the mass center of the body in which the vector from the mass center to the new point would be: `1 * body.frame.x + 2 * body.frame.y + 3 * body.frame.z`.

moorepants:
To clarify some more about the last comment. Adding forces to bodies don't really make sense. Why does a force need to be stored on a body. Does a Body have a force? There are a set of points that are fixed in body's reference frame. But there also exist Points that are not fixed in any bodies' reference frames. And these points can have forces acting on them, the force would contribute to all bodies. One example, is if we have three bodies floating in space. Each body has a spring attached to it. The free end of the springs are tied to together and we specify a force that acts on the combined end of the springs. In that case, the force acts on all bodies and the point it acts on moves.

pydy/forces.py
sahilshekhawat: I just wanted to make things clear for a user so that he by mistake doesn't include the magnitude of direction vector iteself in the force but keeping it simple is also fine.

moorepants: Anytime you create a vector you are specifying its magnitude. So there is no reason to have the magnitude option.

sahilshekhawat: Yup

sahilshekhawat:
I have been trying all of those things that you mentioned here. Thanks a lot,  I don't have a insight into the dynamics part and you are helping me a lot inspite of being busy and I know that I have been slow in pushing code for you guys to review but just want to tell you that I have been working, trying different things and how they might work. I can just push methods but I also think how they might be implemented.

moorepants:
Yeh, I know the dynamics part is hard. It is hard for me too. We'll do our best to give you the right tips on that.

Maybe we should have a discussion about the dynamics. Can you meet on hangouts soon?

And don't be shy about pushing some code up every day, so that we can give you feedback. It doesn't matter the quality as much as it helps show us what you are thinking.

sahilshekhawat:
I am always available, we can have a meeting whenever you are free. Ping me.

I will make sure that I will push at the end of the day so that you guys can review. Thanks :smile:

moorepants:
Here is one option to not dwell on this Force class. If you go with `JointMethod`, then all that is needed is:

```python
jm = JointMethod(...)
jm.loads = seqence_of_force_torque_tuples
```

Don't even worry about this class yet.

moorepants:
I can chat with you in a few minutes.

sahilshekhawat:
Sure

PR #255, opened 2015-07-03
Unit tests for Joint classes

sahilshekhawat:
All the details are in docstrings of fucntions

  commit message.
  directory)
  follow deprecation cycles.)

### DIFF
--- a/pydy/joints.py
+++ b/pydy/joints.py
@@ -1,6 +1,6 @@
 from sympy import acos
-from sympy.physics.vector import cross, dot
-from sympy.physics.mechanics import dynamicssymbols
+from sympy.physics.vector import cross, dot, Vector
+from sympy.physics.mechanics import dynamicsymbols
 
 __all__ = ['Joint', 'PinJoint', 'SlidingJoint', 'CylindricalJoint',
            'SphericalJoint', 'PlanarJoint']
@@ -42,9 +42,10 @@ class Joint(object):
         self.child = child
         self._coordinates = []
         self._speed = []
+        self._kd = []
 
         if parent_point_pos is None:
-            parent_point_pos = (0,0,0)  # Center of mass
+            parent_point_pos = (0, 0, 0)  # Center of mass
         self.parent_point_pos = parent_point_pos
 
         if child_point_pos is None:
@@ -68,16 +69,18 @@ class Joint(object):
         self.parent.child = self.child
 
     def _convert_tuple_to_vector(self, frame, pos_tuple):
-        if len(pos_tuple) == 3:
+        if len(pos_tuple) != 3:
+            raise TypeError('position tuple must be of length 3')
+        else:
             unit_vectors = [frame.x, frame.y, frame.z]
-            vector = 0
+            vector = Vector(0)
             for i in range(3):
                 vector += pos_tuple[i] * unit_vectors[i]
             return vector
-        else:
-            raise TypeError('position tuple must be of length 3')
 
     def _locate_joint_point(self):
+        print "##parent-joint-vector: ", self.parent_joint_vector
+        print "##child-joint-vector: ", self.child_joint_vector
         self.parent_joint_point = self.parent.get_masscenter().locatenew(
             self.name + '_parent_joint',
             self.parent_joint_vector)
@@ -90,6 +93,12 @@ class Joint(object):
         mag2 = vec2.magnitude()
         return acos(dot(vec1, vec2)/(mag1 * mag2))
 
+    def add_kd(self, kd):
+        self._kd.append(kd)
+
+    def get_kds(self):
+        return self._kd
+
     def add_coordinate(self, coordinate):
         self._coordinates.append(coordinate)
 
@@ -99,7 +108,7 @@ class Joint(object):
     def add_speed(self, speed):
         self._speed.append(speed)
 
-    def get_speed(self):
+    def get_speeds(self):
         return self._speed
 
     def apply_joint(self):
@@ -113,36 +122,40 @@ class Joint(object):
 class PinJoint(Joint):
     def __init__(self, name, parent, child, parent_point_pos=None,
                  child_point_pos=None, parent_axis=None, child_axis=None):
-        super(Joint, self).__init__()
 
         if parent_axis is None:
-            self.parent_axis = self.parent.z
+            self.parent_axis = parent.get_frame().x
         else:
             self.parent_axis = parent_axis
 
         if child_axis is None:
-            self.child_axis = self.child.z
+            self.child_axis = child.get_frame().x
         else:
             self.child_axis = child_axis
+
+        super(PinJoint, self).__init__(name, parent, child, parent_point_pos,
+                                       child_point_pos)
 
     def align_axes(self):
         """Rotates child_frame such that child_axis is aligned to parent_axis.
         """
         angle = self.get_angle(self.parent_axis, self.child_axis)
-        self.child.get_frame().orient(
-            'Axis',
-            [angle, cross(self.child_axis, self.parent_axis)])
+        axis = cross(self.child_axis, self.parent_axis)
+        if axis != Vector(0):
+            self.child.get_frame().orient(
+                self.parent.get_frame(), 'Axis', [angle, axis])
 
     def apply_joint(self):
-        self.align_axes()
-        theta = dynamicssymbols('theta')
-        omega = dynamicssymbols('omega')
+        theta = dynamicsymbols('theta')
+        thetad = dynamicsymbols('theta', 1)
+        omega = dynamicsymbols('omega')
         self.add_coordinate(theta)
         self.add_speed(omega)
+        self.add_kd(thetad - omega)
         self.child.get_frame().orient(self.parent.get_frame(), 'Axis',
-                                [theta, self.parent_axis])
+                                      [theta, self.parent_axis])
         self.child.get_frame().set_ang_vel(self.parent.get_frame(), omega * self.parent_axis)
-        self._locate_joint_point()
+        self.align_axes()
         self.child_joint_point.set_pos(self.parent_joint_point, 0)
         self.child.get_masscenter().v2pt_theory(self.parent.get_masscenter(),
                                                 self.parent.get_frame(),
@@ -152,7 +165,8 @@ class PinJoint(Joint):
 class SlidingJoint(Joint):
     def __init__(self, name, parent, child, parent_point_pos, child_point_pos,
                  parent_dir=None, child_dir=None):
-        super(Joint, self).__init__()
+        super(SlidingJoint, self).__init__(name, parent, child, parent_point_pos,
+                                           child_point_pos)
 
         if parent_dir is None:
             self.parent_direction = self.parent.get_frame().z
@@ -174,8 +188,8 @@ class SlidingJoint(Joint):
 
     def apply_joint(self):
         self.align_directions()
-        dis = dynamicssymbols('dis')
-        vel = dynamicssymbols('vel')
+        dis = dynamicsymbols('dis')
+        vel = dynamicsymbols('vel')
         self.add_coordinate(dis)
         self.add_speed(vel)
         self.child.get_frame().orient(self.parent.get_frame(), 'Axis',
@@ -189,27 +203,3 @@ class SlidingJoint(Joint):
                                        vel * self.parent_direction)
         self.child.get_masscenter().set_vel(self.parent.get_frame(),
                                             vel * self.parent_direction)
-
-
-class CylindricJoint(Joint):
-    def __init__(self, name, parent, child, parent_point_pos, child_point_pos):
-        super(Joint, self).__init__()
-
-    def apply_joint(self):
-        # TODO
-
-
-class SphericalJoint(Joint):
-    def __init__(self, name, parent, child, parent_point_pos, child_point_pos):
-        super(Joint, self).__init__()
-
-    def apply_joint(self):
-        # TODO
-
-
-class PlanarJoint(Joint):
-    def __init__(self, name, parent, child, parent_point_pos, child_point_pos):
-        super(Joint, self).__init__()
-
-    def apply_joint(self):
-        # TODO
--- a/pydy/joints.py
+++ b/pydy/joints.py
@@ -150,11 +150,45 @@ class PinJoint(Joint):
 
 
 class SlidingJoint(Joint):
-    def __init__(self, name, parent, child, parent_point_pos, child_point_pos):
+    def __init__(self, name, parent, child, parent_point_pos, child_point_pos,
+                 parent_dir=None, child_dir=None):
         super(Joint, self).__init__()
 
-    def apply_joint(self):
+        if parent_dir is None:
+            self.parent_direction = self.parent.get_frame().z
+        else:
+            self.parent_direction = parent_dir
 
+        if child_dir is None:
+            self.child_direction = self.child.get_frame().z
+        else:
+            self.child_direction = child_dir
+
+    def align_directions(self):
+        """Rotates child_frame such that child_axis is aligned to parent_axis.
+        """
+        angle = self.get_angle(self.parent_direction, self.child_direction)
+        self.child.get_frame().orient(
+            'Axis',
+            [angle, cross(self.child_direction, self.parent_direction)])
+
+    def apply_joint(self):
+        self.align_directions()
+        dis = dynamicssymbols('dis')
+        vel = dynamicssymbols('vel')
+        self.add_coordinate(dis)
+        self.add_speed(vel)
+        self.child.get_frame().orient(self.parent.get_frame(), 'Axis',
+                                      [0, self.parent.get_frame().z])
+        self._locate_joint_point()
+        self.parent_joint_vector.set_vel(self.parent.get_frame(), 0)
+        self.child_joint_point.set_vel(self.child.get_frame(), 0)
+        self.child_joint_point.set_pos(self.parent_joint_point,
+                                       dis * self.parent_direction)
+        self.child_joint_point.set_vel(self.parent.get_frame(),
+                                       vel * self.parent_direction)
+        self.child.get_masscenter().set_vel(self.parent.get_frame(),
+                                            vel * self.parent_direction)
 
 
 class CylindricJoint(Joint):
--- a/pydy/joints.py
+++ b/pydy/joints.py
@@ -1,136 +1,126 @@
-__all__ = ['PinJoint', 'SlidingJoint', 'CylindricalJoint', 'SphericalJoint', 'PlanarJoint']
+__all__ = ['Joint', 'PinJoint', 'SlidingJoint', 'CylindricalJoint',
+           'SphericalJoint', 'PlanarJoint']
 
 
 class Joint(object):
-    """Serves as a base class for all specific Joints. Contains all public methods
-    to be used by other specific Joint's implementation.
+    """Base class for all Joints
 
-    Note: Some of these methods can be shifted to body. Want to have a discussion
-    on this.
+    Joints take two bodies as arguments (one is parent body and other is child
+    body) and define movements of child w.r.t parent. The movement defined is
+    specific to type of Joint. This base class holds all the common and public
+    methods to interact with all Joints. User can also create a custom joint
+    by creating its subclass and overriding 'apply_joint' method. In this method
+    define the movements of child w.r.t to parent.
+
+    Parameters
+    ----------
+    name: String
+        It defines the name of a joint which makes it unique.
+    parent: Body
+        Parent body.
+    child: Body
+        Child body.
+    parent_point_pos: 3 Tuple (optional)
+        Defines the vector to the Joint's point where the parent will be
+        connected to child. 3 Tuple defines the values of x, y and z directions
+        w.r.t parent's frame.
+    child_point_pos: 3 Tuple (optional)
+        Defines the vector to the Joint's point where the child will be
+        connected to parent. 3 Tuple defines the values of x,y and z directions
+        w.r.t child's frame.
+
     """
-    def __init__(self, name=None, parent, par_point_vec_tuple, child, child_point_vec_tuple):
-        if name is None:
-            self._create_name()
-        else:
-            self.name = name
+
+    def __init__(self, name, parent, child, parent_point_pos=None,
+                 child_point_pos=None):
+        self.name = name
         self.parent = parent
         self.child = child
-        self.parent_joint_vector = par_point_vec_tuple
-        self.child_joint_vector = child_point_vec_tuple
-        self.counter = 0
-        self._set_parent_child_rel()
 
-    def _create_name(self, type=None):
-        self.counter += 1
-        if type is None:
-            type = ''
-        self.name = type + 'Joint' + str(self.counter)
+        if parent_point_pos is None:
+            parent_point_pos = (0,0,0)  # Center of mass
+        self.parent_point_pos = parent_point_pos
+
+        if child_point_pos is None:
+            child_point_pos = (0,0,0)  # Center of mass
+        self.child_point_pos = child_point_pos
+
+        self.parent_joint_vector = self._convert_tuple_to_vector(
+            self.parent.frame,
+            self.parent_point_pos)
+        self.child_joint_vector = self._convert_tuple_to_vector(
+            self.child.frame,
+            self.child_point_pos)
+
+        self._set_parent_child_rel()
+        self.apply_joint()
 
     def _set_parent_child_rel(self):
         self.child.parent = self.parent
         self.parent.child = self.child
 
-    def _locate_joint_point(self):
-        """Locates a new point, Point of Joint, using parent's center of mass
-        and parent_joint_vector and sets position of child using child_joint_vector
-        from joint point."""
-        self.joint_point_in_parent = self.parent.masscenter.locatenew(self.name + '_PointInParent', self.)
-        self.joint_point_in_child = self.child.masscenter.locatenew(self.name + '_PointInChild', self.parent)
-
-    def set_joint_point_vel(self, value):
-        """Sets the velocity of the point at joint w.r.t parent's frame."""
-        # TODO
-
-    def set_child_vel(self, value):
-        """sets the velocity of child masscenter to value w.r.t parent's frame.
-        Note that the velocity is added to point where as angular velocity is
-        added to frame."""
-        # TODO
-
-    def set_child_angular_vel(self, value):
-        """sets angular velocity of child's frame w.r.t. to parent's frame.
-        Note that the velocity is added to point where as angular velocity is
-        added to frame."""
-        # TODO
-
-    def orient_child(self):
-        # Note: Can be removed and just orient() be used, including it
-        # here as a reference of available functionalities.
-        """Orients child's frame w.r.t parent's frame. Please refer to
-        sympy.physics.vector.frame.ReferenceFrame.orient() for all the
-        paramter details. Its exactly same"""
-        # TODO
-
-    def _convert_tuple_to_vector(self, frame, tuple):
-        if len(tuple) == 3:
+    def _convert_tuple_to_vector(self, frame, pos_tuple):
+        if len(pos_tuple) == 3:
             unit_vectors = [frame.x, frame.y, frame.z]
             vector = 0
             for i in range(3):
-                vector += tuple[i] * unit_vectors[i]
+                vector += pos_tuple[i] * unit_vectors[i]
             return vector
         else:
-            raise TypeError('tuple must be of length 3')
+            raise TypeError('position tuple must be of length 3')
 
-    def _apply_joint(self):
-        """For custom joint, make a subclass of Joint and override this method."""
-        #TODO generic joint creation.
+    def _locate_joint_point(self):
+        self.parent_joint_point = self.parent.masscenter.locatenew(
+            self.name + '_parent_joint',
+            self.parent_joint_vector)
+        self.child_joint_point = self.child.masscenter.locatenew(
+            self.name + '_child_joint',
+            self.child_joint_vector)
+
+    def apply_joint(self):
+        """To create a custom joint, define a subclass of Joint class and
+        and override this method"""
+        raise NotImplementedError("To define a custom pydy.Joint, you need to" +
+                                  " override apply_joint method in Joint's" +
+                                  " subclass.")
 
 
 class PinJoint(Joint):
-    """Uses methods in Joint's class and create a Revolute (Pin) Joint between
-    parent and child."""
-    def __init__(self, name, parent, child, par_point_vec_tuple=None, child_point_vec_tuple=None, parent_axis=None, child_axis=None):
-        super(Joint, self).__init__(*args, **kwargs)
+    def __init__(self, name, parent, child, parent_point_pos=None,
+                 child_point_pos=None):
+        super(Joint, self).__init__()
+
+    def apply_joint(self):
         # TODO
-
-        self.parent_joint_vector = self._convert_tuple_to_vector(self.parent.frame, par_point_vec_tuple)
-        self.child_joint_vector = self._convert_tuple_to_vector(self.child.frame, child_point_vec_tuple)
-
-        self._apply_joint()
-
-    def _apply_joint(self):
-        """Specific implementation of a joint using all public and private methods."""
-        #TODO
 
 
 class SlidingJoint(Joint):
-    def __init__(self, name, parent, child, par_point_vec_tuple=None,
-                 child_point_vec_tuple=None, direction1=None, direction2=None):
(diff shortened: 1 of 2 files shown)


sahilshekhawat:
All the specific Joint's tests will be in next commit.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,155 @@
...
+
+        # TODO other similar frame.orient() tests can be done.
+        # leaving more tests to discuss its necessity with others.
+
+
+class TestPinJoint():
+    def setup(self):
+        self.parent = Body('parent')
+        self.child = Body('child')
+        self.pinjoint = PinJoint('pinjoint', self.parent, self.child)
```
moorepants: All of our object take a "name" as the first arg. This is likely because the first objects we ever created were mathematical objects. I'd personally rather see the name be optional with a default name being created if you don't pass one. You can make unique default names if you keep a counter on the class, e.g.:

```python
class PinJoint(object):
    counter = 0
    def _create_name(self):
        self.name = something + str(counter)
        counter += 1
```

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def test_pinjoint_init(self):
+        # default values
+        assert self.pinjoint.axis1 == self.parent.frame.x
+        assert self.pinjoint.axis2 == self.child.frame.x
```
moorepants: if you are using the notion of parent and child, maybe you should name these "parent_axis" and "child_axis" so it is more informative.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,155 @@
...
+        assert self.pinjoint.parent_joint_point == self.parent.masscenter
```
moorepants: maybe you can ommit "_joint" in the name. There are only only four points on a PinJoint: parent_point, parent_mass_center, child_point, child_mass_center.

pydy/tests/test_joints.py
sahilshekhawat: introducing a method in each class is probably a good idea . It eliminates the need to build a class for this. Okay, will include this one. thanks

pydy/tests/test_joints.py
sahilshekhawat: yeah, will reduce the name.

pydy/tests/test_joints.py
sahilshekhawat: Acknowledged

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,155 @@
...
+
+        assert self.parent_joint_point.vel(self.parent.frame) == 0
+        assert self.child_joint_point.vel(self.child.frame) == 0
+
+    def test_pinjoint_parameters(self):
+        self.pinjoint = PinJoint('pinjoint', self.parent, self.child,
+                                 (1, 0, 0), (0, 1, 0), axis1='x', axis2='y')
+        assert self.pinjoint.axis1 == self.parent.frame.z
+        assert self.pinjoint.axis2 == self.child.frame.y
+        point1 = self.parent.masscenter.locatenew(self.parent.frame, self.parent.frame.z)
```
moorepants: should this be along the x direction?

pydy/tests/test_joints.py
moorepants: This may be a bad idea though. It may be worth following the pattern of other objects. But I'm thinking that the viz module has mostly optional names.

sahilshekhawat:
If the implementational details a fine, should I push the slidingjoint as well. I have implemented it on the similar ground .

pydy/tests/test_joints.py
sahilshekhawat: I am just pushing the updated code, I changed it in the end and forgot to update it. I have fixed in latest code though.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,155 @@
...
+        point2 = self.child.masscenter.locatenew(self.child.frame, self.child.frame.y)
+        assert self.pinjoint.parent_joint_point == point1
```
moorepants: points are mutable, so this check may not work

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,155 @@
...
+        assert self.pinjoint.child_joint_piont == point2
+        assert self.pinjoint.parent_joint_vector == self.parent.frame.z
+        assert self.pinjoint.child_joint_vector == self.chidl.frame.y
+
+    def test_joint_joint_functions(self):
+
+        # part 1 assining parent-child relationship
+        self.pinjoint.set_parent_child_rel()
```
moorepants: Shouldn't init take care of this? Or will this be a method called in init? It'd should likely be private.

moorepants: Also, if we can avoid it, do not use get/set methods. It is preferable to use the @property decorators.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,155 @@
...
+        assert self.pinjoint.parent_joint_vector == parent_joint_point
+        assert self.pinjoint.child_joint_vector == child_joint_point
+
+        # part 3 aligning axis1 and axis2
+        self.join_frames()
+        # sample implementation
+        # self.child.frame.orient(self.parent.frame, 'Axis', [0, self.parent.frame.x])
+        assert cross(self.pinjoint.axis1, self.pinjoint.axis2) == self.child.frame.z
+        assert cross(self.pinjoint.axis2, self.pinjoint.axis1) == - self.parent.frame.z
+        self.align_axes()
```
moorepants: I would try an exmple joint here and make sure you know what needs to happen. I said that you "align the axes" when we talked, but it is more subtle than that. Aligning axes with the mechanics API, is more like creating the proper ref frame ref frame orientation and creating the right points. It worth doing this manually to make sure you understand the needed operations.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,155 @@
...
+        #                         cross(axis1, axis2)])
+        assert acos(dot(axis1, axis2)/(axis1.magnitude() * axis2.magnitude())) == 0
+
+        # part 4 adding angular velocity to child w.r.t parent
+        assert self.child.frame.ang_vel_in(self.parent.frame) != 0
+
+    def test_joint_apply_joint(self):
+        # apply_joint() should do everything done above by calling specific funtions
+        self.new_pinjoint = PinJoint('pinjoint', self.parent, self.child)
+        self.new_pinjoint._apply_joint()
```
moorepants: Can't all this be called in init?

pydy/tests/test_joints.py
sahilshekhawat: Yes, init will call apply_joint , which will do all of this. here I have tests all methods instead of using apply_joint/init

moorepants:
Sahil, the PinJoint tests are looking great. I think you are getting the hang of it. That's really good start.

pydy/tests/test_joints.py
sahilshekhawat: Just to differentiate between functionalities, I create a method apply_joint which is called by init

pydy/tests/test_joints.py
moorepants: Also, if you do this, this could be something that a Joint abstract base class could contain.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,208 @@
...
+        assert self.pinjoint.child_point == self.child.masscenter
+        assert self.pinjoint.parent_joint_vector == self.parent.frame.x
+        assert self.pinjoint.child_joint_vector == self.child.frame.x
+
+        assert self.parent_point.vel(self.parent.frame) == 0
+        assert self.child_point.vel(self.child.frame) == 0
+
+    def test_pinjoint_parameters(self):
+        self.pinjoint = PinJoint('pinjoint', self.parent, self.child,
+                                 (1, 0, 0), (0, 1, 0), parent_axis='x', child_axis='y')
```
moorepants: You'll also need to write tests for `parent_axis=(a, b, c)` and for child_axis.

pydy/tests/test_joints.py
sahilshekhawat: I have sample implementation which I have tried and it works. You can check if its correct.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def test_joint_orient_child(self):
+        q1 = Symbol('q1')
+        self.joint.orient_child('Axis', [q1, self.parent.frame.x])
+
+        # TODO other similar frame.orient() tests can be done.
+        # leaving more tests to discuss its necessity with others.
+
+
+class TestPinJoint():
```
moorepants: You need some methods/attributes that will generate and expose the single configuration variable for this joint. Every pin joint has one configuration variable (one degree of freedom) that will be crucial for your other methods.

pydy/tests/test_joints.py
sahilshekhawat: Acknowledged.

pydy/tests/test_joints.py
moorepants: Also, please have a look at the `model.n_link_pendulum_with_cart` code. This implements pinjoints with particles instead of rigid bodies. Ideally we support that too.

Maybe your init method should accept RigidBody or Particles. You can also have a pin joint between a particle and a rigidbody. Or you will need to create a Body class that is intialized by passing in a particle or rigidbody.

sahilshekhawat: I think body class should handle that because if we directly accept Particle in Joint then there is a problem of reference frame, user can pass that too and we can also create it. I am pinjoint to accept only pydy.body.Body which can be instantiated from Particle or RigidBody.

moorepants: Ok, that seems fine. I'll be looking forward to the Body unit tests to see how this will work.

sahilshekhawat: I have update Body's unit tests and some sample implementation.

pydy/joints.py
```diff
@@ -0,0 +1,136 @@
...
+    def __init__(self, name=None, parent, par_point_vec_tuple, child, child_point_vec_tuple):
+        if name is None:
+            self._create_name()
+        else:
+            self.name = name
+        self.parent = parent
+        self.child = child
+        self.parent_joint_vector = par_point_vec_tuple
+        self.child_joint_vector = child_point_vec_tuple
+        self.counter = 0
```
moorepants: Unless you make this a class level attribute, you will not ensure unique names in a given Python session, no?

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        self.pinjoint = PinJoint('pinjoint', self.parent, self.child)
+
+    def test_pinjoint_init(self):
+        # default values
+        assert self.pinjoint.parent_axis == self.parent.frame.x
+        assert self.pinjoint.child_axis == self.child.frame.x
+        assert self.pinjoint.parent_point == self.parent.masscenter
+        assert self.pinjoint.child_point == self.child.masscenter
+        assert self.pinjoint.parent_joint_vector == self.parent.frame.x
+        assert self.pinjoint.child_joint_vector == self.child.frame.x
```
moorepants: What is the difference in the parent_joint_vector and the parent_axis? Don't you just only need a vector and point on each body?

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+
+        assert self.parent_point.vel(self.parent.frame) == 0
+        assert self.child_point.vel(self.child.frame) == 0
+
+    def test_pinjoint_parameters(self):
+        self.pinjoint = PinJoint('pinjoint', self.parent, self.child, par_point_vec_tuple=(1, 0, 0),
```
moorepants: I'm not fond of "par_", it should be "parent_" but we could maybe have a shorter name..."parent_point" should be enough. Then in the doc string you can specify that this should be a three tuple.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+                                 child_point_vec_tuple=(0, 1, 0), parent_axis='x', child_axis='y')
+        assert self.pinjoint.parent_axis == self.parent.frame.x
+        assert self.pinjoint.child_axis == self.child.frame.y
+        point1 = self.parent.masscenter.locatenew(self.parent.frame, self.parent.frame.x)
```
moorepants: Isn't the first arg of "locatenew" a string, not a frame?

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        point2 = self.child.masscenter.locatenew(self.child.frame, self.child.frame.y)
+        assert self.pinjoint.parent_point == point1
+        assert self.pinjoint.child_joint_piont == point2
+        assert self.pinjoint.parent_joint_vector == self.parent.frame.x
+        assert self.pinjoint.child_joint_vector == self.chidl.frame.y
```
moorepants: misspell

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        parent_point = parent_masscenter.locatenew(self.name + '_Point', self.parent_joint_vector)
+        child_point = child_masscenter.locatenew(self.name + '_Point', self.child_joint_vector)
+        assert self.pinjoint.parent_joint_vector == parent_point
+        assert self.pinjoint.child_joint_vector == child_point
+
+        # part 3 aligning parent_axis and child_axis
+        self.join_frames()
+        # sample implementation
+        # self.child.frame.orient(self.parent.frame, 'Axis', [0, self.parent.frame.x])
+        assert cross(self.pinjoint.parent_axis, self.pinjoint.child_axis) == self.child.frame.z
```
moorepants: I don't think this is correct. Is this pre-setting the orientation? or or after?

If it is pre-orientation this will just fail with an error because the vectors are not related to each other. Also, I don't see where the axis were set, so how do you know what the cross product would do?

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        self.align_axes()
+        parent_axis = self.pinjoint.parent_axis
+        child_axis = self.pinjoint.child_axis
+        # sample implementation
+        # perpendicular_axis_in_parent = - cross(parent_axis, child_axis)
+        # angle_between_axes = acos(dot(self.pinjoint.child_axis, self.pinjoint.parent_axis))
+        # self.child.frame.orient(self.parent.frame, 'Axis',
+        #                        [acos(dot(parent_axis, child_axis)/(parent_axis.magnitude() * child_axis.magnitude())),
+        #                         cross(parent_axis, child_axis)])
+        assert acos(dot(parent_axis, child_axis)/(parent_axis.magnitude() * child_axis.magnitude())) == 0
```
moorepants: This will probably need to be `assert trigsimp(expr) == 0`.

So here you make sure that the angle between the two vectors is zero after alignment. That's good.

sahilshekhawat: Yup.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+
+        # part 4 adding angular velocity to child w.r.t parent.
+        assert self.child.frame.ang_vel_in(self.parent.frame) != 0
```
moorepants: You should have some checks here to find the configuration variable that specifies the angle.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        point1 = self.parent.masscenter.locatenew(self.parent.frame, self.parent.frame.x)
+        point2 = self.child.masscenter.locatenew(self.child.frame, self.child.frame.y)
+        assert self.pinjoint.parent_point == point1
+        assert self.pinjoint.child_joint_piont == point2
+        assert self.pinjoint.parent_joint_vector == self.parent.frame.x
+        assert self.pinjoint.child_joint_vector == self.chidl.frame.y
+
+    def test_pinjoint_tuple_vector(self):
+        a, b, c, d, e, f = symbols('a b c d e f')
+        pinjoint = PinJoint('pinjoint', self.parent, self.child, par_point_vec_tuple=(a, b, c),
```
moorepants: What if you want a constant angle versus a time varying angle? Or maybe even a consant numerical value for the angle. How will the user specify this? Also what if you want a custom symbol/dynamicsymbol?

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        self.new_pinjoint._apply_joint()
+        # part 1
+        assert self.child.parent == self.parent
+        assert self.parent.child == self.child
+
+        # part 2
+        child_masscenter = self.child.masscenter
+        parent_masscenter = self.parent.masscenter
+        parent_point = parent_masscenter.locatenew(self.name + '_Point', self.parent_joint_vector)
+        child_point = child_masscenter.locatenew(self.name + '_Point', self.child_joint_vector)
```
moorepants: Why do you need to create a new point for the default? Should the default be that the joint is located at the mass center?

pydy/tests/test_joints.py
sahilshekhawat: I thought about that and I have updated my implementation. This is common to other joints e.g. instead of velocity we can set a distance. i will update it after finishing what I am working on.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        assert self.pinjoint.parent_joint_vector == parent_point
+        assert self.pinjoint.child_joint_vector == child_point
+
+        # part 4
+        parent_axis = self.pinjoint.parent_axis
+        child_axis = self.pinjoint.child_axis
+        assert acos(dot(parent_axis, child_axis)/(parent_axis.magnitude() * child_axis.magnitude())) == 0
```
moorepants: It would be nice to have some methods on Vector that implement some basic checks, like:

`vec.aligned(vec2) == True`

pydy/tests/test_joints.py
sahilshekhawat: You are correct. I forgot to add the arguments while defining the new_pinjoint.

pydy/tests/test_joints.py
sahilshekhawat: Yes, they will be very helpful. Should I raise a PR in SymPy?

pydy/joints.py
moorepants: Please add some unit tests for these names.

pydy/tests/test_joints.py
moorepants: Sure.

moorepants:
@pydy/pydy-developers We'd love some help and input on reviewing Sahil's new class ideas.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
+from sympy import Symbol, acos
+from sympy.physics.vector import Point, dot, cross
+from sympy.physics.mechanics import dynamicsymbols
+
+from ..bodies import Body
+from ..joints import Joint, PinJoint, SlidingJoint, CylindricalJoint, SphericalJoint, PlanarJoint
+
+
+class TestJoints():
```
moorepants: Why plural?

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+                                                       self.joint.parent_joint_vector)
+        assert self.joint_point.name == joint_point
+
+        child_masscenter = Point(self.child.name + '_MassCenter')
+        child_masscenter.set_pos(joint_point, self.joint.child_joint_vector)
+        assert self.child.masscenter == child_masscenter
+
+    def test_joint_set_child_vel(self):
+        # Note: this numeric and symbolic value implementation can be applied anywhere in joints
+        # wherever symbol is used e.g. in set_joint_point_vel()
```
moorepants: I'd prefer if all of the tests used symbols. Why do you need to use numbers?

pydy/tests/test_joints.py
moorepants: Can we make names optional?

pydy/tests/test_joints.py
moorepants: You should be able to optionally specify a dynamic symbol, symbol, or float for the angle (not the velocity, you should be computing that behind the scenes).

pydy/tests/test_joints.py
moorepants: Please change it then.

pydy/tests/test_joints.py
moorepants: Ok, please update the test.

pydy/tests/test_joints.py
moorepants: If you implement this on vector, you need to change this test to use the method.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def test_slidingjoint_init(self):
+        # Default values
+        assert self.slidingjoint.parent == self.parent
+        assert self.slidingjoint.child == self.child
+        assert self.slidingjoint.direction1 == self.parent.x
+        assert self.slidingjoint.direction2 == self.child.x
+        assert self.slidingjoint.parent_point == self.parent.masscenter
+        assert self.slidingjoint.child_point == self.child.masscenter
+
```
moorepants: If these attributes are public can we change them after `__ini__`? This is possible in the system class, for example. It is a nice feature to have.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+    def test_slidingjoint_paramters(self):
```
moorepants: spelling

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        # custom parameters
+        self.slidingjoint = SlidingJoint('slidingjoint', self.parent, self.child, par_point_vec_tuple=(1,0,0),
+                                         child_point_vec_tuple=(0,1,0), direction1='x', direction2='y')
```
moorepants: You need to add a unit test for arbitrary directions.

Also please test using symbols with the `child_point_vec_tuple`.

I think `child_point` is better than `child_point_vec_tuple`. The docstring should explain that it is a three tuple.

pydy/tests/test_joints.py
moorepants: You need some checks for errors if the user passes in incorrect values to init.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+
+        # part 4
+        parent_axis = self.pinjoint.parent_axis
+        child_axis = self.pinjoint.child_axis
+        assert acos(dot(parent_axis, child_axis)/(parent_axis.magnitude() * child_axis.magnitude())) == 0
+
+
+class TestSlidingJoint():
+    def setup(self):
+        self.parent = Body('parent')
```
moorepants: How can a body be initialized with only a string? This should be a complete initialization, no?

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        assert self.slidingjoint.child_point == self.child.masscenter
+
+    def test_slidingjoint_paramters(self):
+        # custom parameters
+        self.slidingjoint = SlidingJoint('slidingjoint', self.parent, self.child, par_point_vec_tuple=(1,0,0),
+                                         child_point_vec_tuple=(0,1,0), direction1='x', direction2='y')
+        assert self.pinjoint.direction1 == self.parent.frame.x
+        assert self.pinjoint.direction2 == self.child.frame.y
+        point1 = self.parent.masscenter.locatenew(self.parent.frame, self.parent.frame.x)
+        point2 = self.child.masscenter.locatenew(self.child.frame, self.child.frame.y)
```
moorepants: I thought these new points should be using the `point_vec_tuple`s?

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        assert self.pinjoint.parent_point == point1
+        assert self.pinjoint.child_joint_piont == point2
+        assert self.pinjoint.parent_joint_vector == self.parent.frame.x
+        assert self.pinjoint.child_joint_vector == self.child.frame.y
+
+    def test_slidingjoint_functions(self):
+        self.slidingjoint._set_parent_child_rel()
```
moorepants: Why is this method needed, can't you just set the parent and child in `__init__`?

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        assert self.parent.child == self.child
+        assert self.child.parent == self.parent
+
+        self.slidingjoint._locate_joint_point()
+        child_masscenter = self.child.masscenter
+        parent_masscenter = self.parent.masscenter
+        parent_point = parent_masscenter.locatenew(self.name + '_Point', self.parent_joint_vector)
+        child_point = child_masscenter.locatenew(self.name + '_Point', self.child_joint_vector)
```
moorepants: What is the `child_joint_vector`? The vector the sliding happens along or the vector to the point the vector lies on?

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        assert self.pinjoint.parent_joint_vector == parent_point
+        assert self.pinjoint.child_joint_vector == child_point
+
+        self.join_frames()
```
moorepants: Does this happen before or after alignment?

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        assert cross(self.pinjoint.parent_axis, self.pinjoint.child_axis) == self.child.frame.z
+        assert cross(self.pinjoint.child_axis, self.pinjoint.parent_axis) == - self.parent.frame.z
+
+        self.align_axes()
+        parent_axis = self.pinjoint.parent_axis
+        child_axis = self.pinjoint.child_axis
+        assert acos(dot(parent_axis, child_axis)/(parent_axis.magnitude() * child_axis.magnitude())) == 0
+
+        # Similar to Pinjoint but have a velocity instead of angular velocity.
+        assert self.child_point.vel(self.parent.frame) != 0
```
moorepants: This test is missing any all information about the velocities. You need tests that make sure the coordinate and velocity associated with this joint have been created properly.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+
+class TestCylindricalJoint():
+    # Note all the functionalitites are similar to other PinJoint and SlidingJoint.Only CylindricalJoint's specific
+    # tests are written.
+    def setup(self):
+        self.parent = Body('parent')
+        self.child = Body('child')
+        self.cylindricaljoint = CylindricalJoint('cylindricaljoint', self.parent, self.child, par_point_vec_tuple=(1,1,1),
+                                                 child_point_vec_tuple=(0,0,1), parent_direction='x + 2*y', child_direction = 'z')
```
moorepants: Please use only symbols in the tests.

What is `x + 2 * y`? This is not something we likely want to support. The user should be able to supply either `x`, `y`, `z` or a 3-tuple representing the measure numbers of the vector in the parent or child reference frame.

moorepants: Also, this line is too long. Please stick to pep8.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        assert self.cylindricaljoint.child_joint_vector == self.child.frame.z
+        assert self.cylindricaljoint.parent_direction == self.parent.frame.x + 2*self.parent.frame.y
+        assert self.cylindricaljoint.child_direction == self.child.frame.z
+
+    def test_cylindricaljoint_apply_joint(self):
+        assert self.child_point.vel(self.parent.frame) != 0
+        parent_axis = self.pinjoint.parent_direction
+        child_axis = self.pinjoint.child_direction
+        assert acos(dot(parent_axis, child_axis)/(parent_axis.magnitude() * child_axis.magnitude())) == 0
+
```
moorepants: This test is missing all info about the two coordinates and the two velocities associated with this joint.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+
+class TestSphericalJoint():
+    def setup(self):
+        self.parent = Body('parent')
+        self.child = Body('child')
+        self.sphericaljoint = SphericalJoint(self.parent, self.child, par_point_vec_tuple=(0,2,0),
+                                             child_point_vec_tuple=(0,1,1), parent_plane='x*y', child_plane='y*z + 2*z*x')
```
moorepants: Why does a spherical joint need a plane? Should two points be sufficient?

moorepants: Please write tests with symbols not numbers.

moorepants: This is not a good api design:

parent_plane='x*y', child_plane='y*z + 2*z*x'

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+                                             # yz is y-z plane
+        # Note: Refer to SphericalJoint's docstring for information about implementation using planes.
+
+    def test_spehricaljoint_paramters_assignment(self):
+        assert self.sphericaljoint.parent == self.parent
+        assert self.sphericaljoint.child == self.child
+        assert self.sphericaljoint.parent_joint_vector == 2 * self.parent.frame.y
+        assert self.sphericaljoint.child_joint_vector == self.child.frame.y + self.child.frame.z
+        assert self.sphericaljoint.parent_plane == # TODO
+        assert self.sphericaljoint.child_plane == # TODO
```
moorepants: This test is missing all info about coordinates and velocities. It should use quaternions as the base implementation. So there will be four coordinates and four velocities. Try making a spherical joint manually with quaternions and then write the tests that will exercise that.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+
+
+class PlanarJoint():
+    def setup(self):
+        self.parent = Body('parent')
+        self.child = Body('child')
+        self.planarjoint = PlanarJoint(self.parent, self.child, par_point_vec_tuple=(0,1,0),
+                                       child_point_vec_tuple=(1,0,0), parent_plane='xy', child_plance='yz + zx')
```
moorepants: Not acceptable:  parent_plane='xy', child_plance='yz + zx'

This should take a distance between the planes and normal vectors to the plane.

pydy/tests/test_joints.py
```diff
@@ -0,0 +1,275 @@
...
+        # Note: Refer to SphericalJoint's docstring for information about implementation using planes.
+
+    def test_planarjoint_parameters_assignment(self):
+        assert self.planarjoint.parent == self.parent
+        assert self.planarjoint.child == self.child
+        assert self.planarjoint.parent_joint_vector == self.parent.frame.y
+        assert self.planarjoint.child_joint_vector == self.child.frame.x
+        assert self.planarjoint.parent_plane == # TODO
+        assert self.planarjoint.child_plance == # TODO
```
moorepants: This test is far from complete.

moorepants:
Sahil, I just reviewed the code. Your tests are far from complete. You have no concepts of coordinates and velocities.

Please try to make each type of joint manually with the sympy vector/mechanics api and you will see that you are missing a lot of stuff.

All of the joints should supply the relative kinematic differential equation too.

Here is an example PinJoint:

```python
In [1]: import sympy as sm

In [2]: from sympy.physics.vector import *

In [3]: parent_frame = ReferenceFrame('parent')

In [4]: child_frame = ReferenceFrame('child')

In [5]: parent_mass_center = Point('parent_com')

In [6]: child_mass_center = Point('child_com')

In [7]: coordinate = dynamicsymbols('theta')

In [9]: parent_joint_point = Point('parent_joint')

In [10]: child_joint_point = Point('child_joint')

In [11]: child_frame.orient(parent_frame, 'Axis', [coordinate, parent_frame.x])

In [14]: omega = dynamicsymbols('omega')

In [16]: child_frame.set_ang_vel(parent_frame, omega * parent_frame.x)

In [17]: child_frame.ang_vel_in(parent_frame)
Out[17]: omega*parent.x

In [18]: a, b, c, d, e, f = sm.symbols('a, b, c, d, e, f')

In [19]: child_joint_loc = (a, b, c)

In [20]: parent_joint_loc = (d, e, f)

In [23]: parent_joint_point.set_pos(parent_mass_center, d * parent_frame.x + e * parent_frame.y + f * parent_frame.z)

In [24]: child_mass_center.set_pos(parent_mass_center, 0)

In [22]: child_joint_point.set_pos(child_mass_center, a * child_frame.x + b * child_frame.y + c * child_frame.z)
```

All of that (and maybe some more) should happen inside of PinJoint.
parent = ReferenceFrame('

sahilshekhawat:
I have updated PInJoint and slidingJoint's implementation based on examples pushed in the separate PR.

moorepants:
This has been submitted to SymPy. Closing.

PR #258, opened 2015-07-23
Examples for Joints

sahilshekhawat:
Will include examples of creating system in two ways:
1. by using Joints implemented in other PRs.
2. by using current API to show how Joints will be implemented.

### DIFF
--- a/examples/pinjoint.py
+++ b/examples/pinjoint.py
@@ -1,0 +1,94 @@
+import sympy as sm
+from sympy.physics.vector import *
+from sympy.physics.mechanics import *
+from numpy import linspace
+from pydy.system import System
+from numpy import pi
+from pydy.viz.shapes import Cylinder, Sphere
+from pydy.viz.scene import Scene
+from pydy.viz.visualization_frame import VisualizationFrame
+
+# Body----------------------------------------------
+parent_masscenter = Point('parent_masscenter')
+parent_frame = ReferenceFrame('parent_frame')
+parent_mass = sm.Symbol('parent_mass')
+
+parent_masscenter.set_vel(parent_frame, 0)
+parent = Particle('parent', parent_masscenter, parent_mass)
+
+child_masscenter = Point('child_masscenter')
+child_frame = ReferenceFrame('child_frame')
+child_mass = sm.Symbol('child_mass')
+
+child_masscenter.set_vel(child_frame, 0)
+child = Particle('child', child_masscenter, child_mass)
+
+print "Finished creating Bodies"
+# Joint----------------------------------------------
+
+# join parent and child frames
+child_frame.orient(parent_frame, 'Axis', [0, parent_frame.x])
+
+coordinate = dynamicsymbols('theta')
+coordinated = dynamicsymbols('theta', 1)
+omega = dynamicsymbols('omega')
+
+a, b, c, d, e, f, g = sm.symbols('a b c d e f g')
+
+child_joint_loc = (a, b, c)
+parent_joint_loc = (d, e, f)
+
+parent_joint_point = parent.get_point().locatenew(
+    'parent_joint',
+    a * parent_frame.x + b * parent_frame.y + c * parent_frame.z)
+
+child_joint_point = child.get_point().locatenew(
+    'child_joint',
+    e * child_frame.x + d * child_frame.y + f * child_frame.z)
+
+child_joint_point.set_pos(parent_joint_point, 0)
+
+
+child_frame.orient(parent_frame, 'Axis', [coordinate, parent_frame.x])
+child_frame.set_ang_vel(parent_frame, omega * parent_frame.x)
+
+child.get_point().v2pt_theory(parent.get_point(), parent_frame, child_frame)
+
+print "Finished creating Joints"
+# JointsMethod----------------------------------------------------
+BL = [parent, child]
+FL = [(parent.get_point(), parent.get_mass() * g * parent_frame.x), (child.get_point(), child.get_mass() * g * parent_frame.x)]
+kd = [coordinated - omega]
+
+KM = KanesMethod(parent_frame, q_ind=[coordinate], u_ind=[omega], kd_eqs=kd)
+(fr, frstar) = KM.kanes_equations(FL, BL)
+print "###########", fr
+print "###########", frstar
+
+print "Finished JointsMethod"
+# Numeric part---------------------------------------------------
+constants = {parent_mass: 1.0, child_mass: 1.0, a: 1.0, b: 1.0, c: 1.0, d: 1.0, e: 1.0, f: 1.0, g: 9.8}
+#constants = {g: 9.8}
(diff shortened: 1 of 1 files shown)


moorepants:
This is going to need to be in a directory for this particular example.

What is the example? Is a two mass chain link?

sahilshekhawat:
I have update slidingjoint and added visualization to both the examples. each file contains entire code to generate visualization.

examples/Joints/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'sahil'
```
moorepants: What's this for?

pydy/system.py
```diff
@@ -609,4 +609,5 @@ def _Kane_constant_symbols(self):
             constants = set(self.eom_method._find_othersymbols(
                 *self._Kane_inlist_insyms()))
         constants.remove(dynamicsymbols._t)
+        print constants
```
moorepants: ?

examples/Joints/__init__.py
sahilshekhawat: My IDE automatically adds this, forgot to remove it.

pydy/system.py
sahilshekhawat: This one too. Fixing now

examples/Joints/cylindricaljoint.py
```diff
@@ -0,0 +1,150 @@
+from sympy import symbols, Symbol, acos
```
sahilshekhawat: This example is WIP

examples/Joints/Ingen/pinjoint.py
```diff
@@ -0,0 +1,16 @@
+from sympy import Symbol
```
sahilshekhawat: This is the same example as simple_pinjoint but using Ingen's API. It gives the same equations of motion.

examples/Joints/Ingen/pinjoint.py
```diff
@@ -0,0 +1,16 @@
...
+from pydy.bodies import Body
+from pydy.joints import PinJoint
+from pydy.joints_method import JointsMethod
+
+parent = Body('parent')
+child = Body('child')
+
+gravity = Symbol('gravity')
+child.add_force((0, child.get_mass() * gravity, 0), child.get_masscenter())
```
moorepants: If you add this to the child how do you know which direction to apply the force in the parent frame?

Let's avoid getter and setter methods. Anytime you have a get/set method you should simply make use of the `@property` decorator to control setting attributes. A method is also only needed if you need to check or change things on attribute change.

examples/Joints/Ingen/pinjoint.py
```diff
@@ -0,0 +1,16 @@
...
+
+pin_joint = PinJoint('pin_joint', parent, child, parent_point_pos=(1, 1, 1),
+                     child_point_pos=(1, 1, 1))
```
moorepants: What are you creating here? Why wouldn't you leave the parent position at (0, 0, 0) if this is a simply pendulum bob. Maybe change this to have the child point be `(0, l, 0)` or `(0, 0, 1)` (depending on which way gravity is.

examples/Joints/Ingen/pinjoint.py
```diff
@@ -0,0 +1,16 @@
...
+
+JM = JointsMethod([pin_joint], parent)
+print JM.get_kanes()
```
moorepants: Avoid "get" methods.

moorepants: Also I don't think the user should be accessing Kane's anything. You should return the mass_matrix, forcing, mass_matrix_full, and forcing_full. Just like the KanesMethod and LagrangesMethod class do.

examples/Joints/Ingen/pinjoint.py
moorepants: Cool!

examples/Joints/Ingen/pinjoint.py
sahilshekhawat: 1. the first argument is a 3 Tuple. Therefore, (a, b, c) == a*frame.x + b*frame.y + c*frame.z

2. force is stored in a list thus, add_force was needed. It appends to the force_list.

examples/Joints/Ingen/slidingjoint.py
```diff
@@ -0,0 +1,21 @@
+from sympy import Symbol
```
moorepants: What are you creating here? Is it a single mass on a spring that has gravity in the same direction as the mass degree of freedom?

examples/Joints/Ingen/pinjoint.py
sahilshekhawat: I agree, it can be changed to anything. I just choose (1, 1, 1)  at the time of writing.

examples/Joints/Ingen/slidingjoint.py
moorepants: Please don't use capital letters in the directory names.

examples/Joints/Ingen/pinjoint.py
sahilshekhawat: I will change it.

examples/Joints/Ingen/pinjoint.py
sahilshekhawat: Got it. Thanks changing it now.

examples/Joints/Ingen/slidingjoint.py
sahilshekhawat: Yes, its a mass and bob example with gravity in the same direction as DoF.

Sorry, I will take care of that next time.

examples/Joints/Ingen/pinjoint.py
moorepants: 1. But adding the force to the child's y direction is incorrect.
2. add_force is not needed, because if `.forces` is just a list, then all you need to do is `.forces.append(...)`

`get_masscenter` should use @property instead.

examples/Joints/Ingen/pinjoint.py
moorepants: I think it would be best to try to create typical systems. Like this one should create a simple pendulum. You can get some ideas from the tests from KanesMethod and LagrangesMethod.

moorepants:
@sahilshekhawat I think most of these examples really need to be unit tests. You can make examples and unit tests, but these super basic examples should be used as tests, much like the other Methods classes.

sahilshekhawat:
Yes, I am using them are unit tests. Please refer to my PR #255

examples/Joints/Ingen/pinjoint.py
sahilshekhawat: get_masscenter is from RigidBody. I have not changed anything there. but Yeah, all the attributes in RigidBody are private and uses setters/getters. Should we change them?

sahilshekhawat: and I cannot do forces.append since I need to convert tuple to vector

PR #259, opened 2015-07-24
[WIP] The start of 3D example which has all base joint types.

moorepants:
@sahilshekhawat This is an example I came up with which has all of your base joint types. I haven't finished it yet but the base structure is there. I'll peck at it more this weekend as I have time. Do you think this will be a good example to use to test your new classes? I figured we could make a single example with all joint types instead of several examples, one for each joint.

Let me know what you think.


sahilshekhawat:
Yes, using all joitns together will be a good to test all components of my API together. Also, the tests that I have can be used for individual joints.

moorepants:
@sahilshekhawat This example works now but still needs some polish. Your joints methods should be able to handle this example and it should exercise most of the functionality of your new classes. It could be used as a basis of a unit test to ensure you get the same equations of motion.

One thing I realized while working on this is that a generic non-singular spherical joint will be some work to implement because you will get singular mass matrices from the quaternions. You can implement a general spherical joint with body fixed coordinates but you may need to figure out how to ensure that the best ordering of the body fixed rotations is chosen by the user for the problem at hand.

sahilshekhawat:
Jason,
I looked to simbody and they mention that 
>Three mobilities – unrestricted orientation modeled with a quaternion which is never singular.

I have completed all other joints and now working on the prototype of spherical joint. I am learning more about things like quaternion etc. to get more idea for the implementation.

PR #264, opened 2015-10-06
Don't assume user enters input in server shutdown

oliverlee:


moorepants:
+1 if the Py 2.7 test passes on the retry. Looks like the numpy install was boogered or something.

moorepants:
This may be the Travis issue: https://github.com/conda/conda/issues/1621

moorepants:
This is failing because of an issue from anaconda's side:

https://github.com/conda/conda/issues/444

They introduced a dependency on a new libfortran package but didn't make it backwards compatible with all NumPy versions. If we switch to NumPy 1.7.1 it should work.

oliverlee:
@moorepants can you merge this?

Issue #32, opened 2014-04-10
We should be able to speed up the evaluation of the transformation matrices with this.


moorepants:
_Issue by **[moorepants](https://github.com/moorepants)** from Sunday Jan 26, 2014 at 03:28 GMT_
_Originally opened as https://github.com/pydy/pydy-viz/issues/97_

----

See the end of this issue:

https://github.com/sympy/sympy/issues/2826

If you overwrite the ImmutableMatrix function with np.array, then you can use vectorized operations instead of a for loop.

oliverlee:
In the method you suggest, constant entries in the transform result in a scalar being returned instead of a numpy ndarray. If a constant matrix entry is evaluated before a varying entry, the evaluation of the lambdified function fails. You can see this by setting ```mat_expr[0, 0] = 0``` in your example.

oliverlee:
Resolved here: https://github.com/pydy/pydy/commit/42aae29ab26d0cbbbfd657c33f32bbd5dc53bb6f

PR #266, opened 2015-10-07
Use vectorized operations to compute transformations

oliverlee:
Use vectorized Numpy operation to compute visualization frame
transformation matrices as this is much faster than using a for loop.
Each element of the transformation matrix is lambdified and stored in a
list of numerical transformations.

Individual element transforms that do not contain dynamic variables nor
constant variables are not returned as functions by lambdify and
subsequent evaluation for these elements is skipped. Element transforms
that contain constant variables but not dynamic variables are lambdified
with only the constant variables as arguments. These are only evaluated
once and then repeated for create of the visualization matrix
trajectory.

This commit resolves issue https://github.com/pydy/pydy/issues/32

### DIFF
--- a/pydy/viz/visualization_frame.py
+++ b/pydy/viz/visualization_frame.py
@@ -225,13 +225,10 @@ class VisualizationFrame(object):
         """
         rotation_matrix = self.reference_frame.dcm(reference_frame)
         self._transform = Identity(4).as_mutable()
-        self._transform[0:3, 0:3] = rotation_matrix[0:3, 0:3]
+        self._transform[0:3, 0:3] = rotation_matrix
 
-        _point_vector = self.origin.pos_from(point).express(reference_frame)
-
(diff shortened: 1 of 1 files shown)


moorepants:
Are there any tests for this code?

moorepants:
Please update with master.

oliverlee:
It looks like there are no tests for the calculation of the transform matrices.

moorepants:
Would you be willing to create a regression test? You could run the code from master on an example and pull out some values and then make the test against them.

oliverlee:
No problem I'm already doing that now.

oliverlee:
Actually, this is already a regression test:
https://github.com/pydy/pydy/blob/d57dac0ff33722c05cdc97b20b60b6743592604e/pydy/viz/tests/test_visualization_frame_scene.py#L177-L212

so I think we can merge this unless you have any other issues.

moorepants:
Sweet. Merging.

PR #268, opened 2015-10-11
Making PyDy compatible with Jupyter(IPython4+)

tarzzz:
This PR is to update the code to be compatible with the upcoming versions of IPython-Jupyter.
Below is the checklist on the things broken/need to modify on upgrading Ipython to Jupyter.
@moorepants Please feel free to add more as per your observations.

  - [ ] Fix Re-run Simulation Button (Takes up the whole width of the canvas)
  - [ ] Markdown does not seem to be working on Jupyter.

### DIFF
--- a/examples/mass_spring_damper/mass_spring_damper.ipynb
+++ b/examples/mass_spring_damper/mass_spring_damper.ipynb
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "#Defining the Problem"
+    "#Defining the Problem#"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


moorepants:
I think the main issue is that ipywidgets is a new package and we need to specify that it is a dependency for IPython >=4.0. Everything ran fine for me with ipywidgets installed.

moorepants:
Please update with master.

tarzzz:
Different modules like 'ipywidgets' etc. are currently maintained as shims in IPython4. They might decide to remove them in future releases, which would cause our code to break.

moorepants:
Yes that's true but for IPython 3 and 4 that isn't the case. When IPython 5 comes out we will have to modify things to be compatible.

tarzzz:
I feel that since it is working (although with a ShimWarning) there is no need to push this code, since it effectively does nothing. Moreover the issue in the first place was with the IPython. We can close this one.

oliverlee:
Please remember to remove your closed and abandoned branches since you are creating them in the PyDy repository: https://github.com/pydy/pydy/branches

moorepants:
@tarzzz Also, please only create branches from now on in your fork on github. We've been trying to follow sympy's method here now that we have more contributors.

tarzzz:
Yep, Got it. I need some time to view the old branches, to confirm they are
not relevant now. Will start reviewing and removing them real soon.

Issue #254, opened 2015-06-30
pydy-resources directory is not relocatable


moorepants:
I tried copying the pydy-resources dir to my server. If you load a scene with `?load=` it errors out because it can't find the simulation file. It looks for the simulation file in the root of the website instead of the directory where index.html is.

tarzzz:
This is related to making pydy-resources directory portable, so that they can simply be copied to any system, and should be able to work good.
#269

moorepants:
I'm using apache as my web server. I'd like a way to export all the files needed to display the simulation on the web, i.e. copy a single directory to a directory that is served by apache (or nginx).

tarzzz:
I am unable to re-produce the issue you are facing @moorepants. Have tried several combinations on my local machine, by creating local servers from different directories. Works fine on them.

tarzzz:
So I have been able to replicate the issue now. Javascript code needs to be modified to read the url, and then infer the simulation data file from the url (when accessing from outside IPython)

tarzzz:
@moorepants Could you please Pull #269 and see if it works for you, both on local as well as on your remote apache server? I have attempted a fix, and want to test it.

moorepants:
Yes, I'll check it on my server. I think it has to do with leading slashes in the urls or something. When you place the files on a server to be served at an arbitrary url, it wasn't working. I'll try soon. Thanks for working on this.

PR #269, opened 2015-10-11
Making PyDy Resources portable.

tarzzz:
Aim is that we should be able to copy the pydy-resources directory to any system, and it should be able to visualize the model it was generated for.


tarzzz:
Modified js code. Should fix #254. Need to study the impact on ipython notebooks though.

moorepants:
Please update with master so the builds stop failing.

PR #270, opened 2015-10-12
Switched the Travis conda installs to latest minor versions.

moorepants:
@oliverlee I think this should fix the libfortran error in issue #264.


moorepants:
Ok, this fixes the travis builds. @oliverlee Please merge if you are fine with this and then you can update your PRs with this change so that the builds pass.

moorepants:
Merging this, so we can have working builds.

PR #272, opened 2015-10-15
Fix scene simulation time

oliverlee:
The time shows up as 0 when the model is loaded (not the first value in the times array). Maybe one of you knows how to fix it since I don't know JS.

We should probably implement some sort of speedup slider to change fps in the future.

### DIFF
--- a/pydy/viz/scene.py
+++ b/pydy/viz/scene.py
@@ -408,15 +408,23 @@ class Scene(object):
         self._generate_scene_dict()
 
         self._scene_info["simulationData"] = self._simulation_json_file
+
+        times = None
         if self.times is not None:
-            # Assume that times is evenly spaced and monotonic.
-            # TODO: Interpolate if times are not evenly spaced.
-            total_time = self.times[-1] - self.times[0]
-            self._scene_info["timeDelta"] = total_time / (num_time_steps - 1)
-            self._scene_info["startTime"] = self.times[0]
-        else:
+            times = self.times
+        elif self.system.times is not None:
+            times = self.system.times
+
+        if times is None:
             self._scene_info["timeDelta"] = 1.0 / self.frames_per_second
             self._scene_info["startTime"] = 0.0
+        else:
+            # Assume that times is evenly spaced and monotonic.
+            # TODO: Interpolate if times are not evenly spaced.
+            total_time = times[-1] - times[0]
+            self._scene_info["timeDelta"] = total_time / (num_time_steps - 1)
+            self._scene_info["startTime"] = times[0]
+
         self._scene_info["fps"] = self.frames_per_second
         self._scene_info["speedup"] = (self.frames_per_second *
                                        self._scene_info["timeDelta"])
--- a/pydy/viz/scene.py
+++ b/pydy/viz/scene.py
@@ -408,9 +408,18 @@ class Scene(object):
         self._generate_scene_dict()
 
         self._scene_info["simulationData"] = self._simulation_json_file
-        # NOTE : Python 3 division is imported at the top of the file so
-        # this will be a float.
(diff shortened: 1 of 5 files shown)


tarzzz:
Its been some time, but as I recall, there was no provision of providing the time vector to javascript. It took only (start_time, end_time and time_delta) and calculated time vector based on these. We will require some changes in code to accomodate both.

oliverlee:
The start time is displayed correctly after the first time the simulation is run. If looped, the start time is correct if non zero.

It is only in the beginning, before the simulation has been run the first time,  where the time is (possibly) incorrectly displayed as 0.

oliverlee:
@tarzzz The time value is hardcoded to zero here: https://github.com/pydy/pydy/blob/eaf45c65f9e2cdeb46dd0ef6c41867e144281376/pydy/viz/static/index.html#L91

Is there a way to set the initial time immediately after loading the model?

oliverlee:
Also, when I modify .js files in the static directory, changes are not reflected when I start a server (or at least it doesn't show the changes when I 'inspect element' in my browser). Is there a way to force this to refresh?

tarzzz:
We will need to set the initial time, dynamically on load. 
Regarding the js files, after you change the code in static directory, you need to re-execute the python code for simulation. 

Another way is to go to pydy-resources directory which is created for that simulation. You can edit code inside the js/ directory, test it there and if it works, change the same code in static directory, and push to github.

oliverlee:
I've deleted the whole pydy resources directory and re-executed all the
python code but there's no change. This is the script I use:
https://github.com/oliverlee/bicycle/blob/master/python/scripts/create_viz.py

I'll try modifying the generated js directory later.

moorepants:
Maybe we need to create a jupyter extension so that the js files for the viz are served from the user's notebook extension directory in their home directory. That could possibly avoid the copying of the files to the working directory.

oliverlee:
This now works for me.

pydy/viz/scene.py
```diff
@@ -408,9 +408,17 @@ def _generate_json(self, directory=None, prefix=None):
...
-        # this will be a float.
-        self._scene_info["timeDelta"] = 1 / self.frames_per_second
+        if self.times is not None:
+            # Assume that times is evenly spaced and monotonic.
+            # TODO: Interpolate if times are not evenly spaced.
+            total_time = self.times[-1] - self.times[0]
+            self._scene_info["timeDelta"] = total_time / (num_time_steps - 1)
+            self._scene_info["startTime"] = self.times[0]
+        else:
+            self._scene_info["timeDelta"] = 1.0 / self.frames_per_second
```
tarzzz: @oliverlee You should include the "startTime" dict key here also (can be set to zero). Else the scene_desc json files will become in-consistent for two cases, and js code will break. Then it should be okay to be merged in.

oliverlee: fixed

oliverlee: This is ready to be merged.

moorepants:
Update with master please.

moorepants:
I just tried this branch with the mass_spring_damper notebook and changes the sys.times var to go from 5.0 to 10.0 seconds with 100 steps in time, but the slider bar gives me 0.0 to 3.3 seconds with 100 steps. The value to the left reflects the seconds and the value in the black box that follows the slider when dragged seems to reflect the frame number. So it doesn't seem to work for me.

tarzzz:
We should come up with a standardized test for checking js related fixes. Fixing one thing might lead to something else breaking, and it is difficult to track unless someone manually runs the simulation and tests it.

moorepants:
Ok, this now works for me. +1.

You need to update with master it seems.

Issue #167, opened 2015-05-14
Add Bechmark's run to travis


sahilshekhawat:
I tried before but it always leads to error
```
$ python bin/benchmark_pydy_code_gen.py 2 4
Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/bin/benchmark_pydy_code_gen.py", line 4, in <module>
    __import__('pkg_resources').run_script('pydy==0.3.0.dev0', 'benchmark_pydy_code_gen.py')
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 499, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 1235, in run_script
    execfile(script_filename, namespace, namespace)
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/lib/python2.7/site-packages/pydy-0.3.0.dev0-py2.7.egg/EGG-INFO/scripts/benchmark_pydy_code_gen.py", line 9, in <module>
    from pydy.models import n_link_pendulum_on_cart
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/pydy-0.3.0.dev0-py2.7.egg/pydy/models.py", line 12, in <module>
    from .system import System
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/pydy-0.3.0.dev0-py2.7.egg/pydy/system.py", line 58, in <module>
    from .codegen.ode_function_generators import generate_ode_function
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/pydy-0.3.0.dev0-py2.7.egg/pydy/codegen/ode_function_generators.py", line 16, in <module>
    from .cython_code import CythonMatrixGenerator
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/pydy-0.3.0.dev0-py2.7.egg/pydy/codegen/cython_code.py", line 11, in <module>
    from .c_code import CMatrixGenerator
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/pydy-0.3.0.dev0-py2.7.egg/pydy/codegen/c_code.py", line 13, in <module>
    from ..utils import wrap_and_indent, find_dynamicsymbols
[...]

PR #277, opened 2015-10-17
Require benchmark to run in Travis

oliverlee:
This commit resolves https://github.com/pydy/pydy/issues/167.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,3 +52,5 @@ script:
   - cd docs && make html && cd ..
   # Make sure the library installs.
   - python setup.py install
+  # Run benchmark
+  - python bin/benchmark_pydy_code_gen.py 1 10
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ install:
       conda install numpy scipy cython theano;
       pip install https://github.com/sympy/sympy/archive/master.zip;
     fi
+  - conda install matplotlib
 before_script:
   - conda info
   - conda list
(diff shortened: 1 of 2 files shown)


bin/benchmark_pydy_code_gen.py
```diff
@@ -5,6 +5,8 @@
 
 # external libraries
 from numpy import hstack, ones, pi, linspace, array, zeros, zeros_like
+import matplotlib
+matplotlib.use('Agg')
```
moorepants: Should this be "if on Travis use agg"?

oliverlee: The benchmark doesn't call plt.show() and just saves the file so it's fine
as is. Works on my machine as is.

We can add the Travis check but I don't know the variable to check.

moorepants: But some people may not have the agg backend install on their on machine (not sure about this, maybe agg is always available on all platforms).

You can check the `TRAVIS` env variable to see if true.

oliverlee: It looks like it is always available:
http://matplotlib.1069221.n5.nabble.com/Any-available-backend-for-non-root-users-td38528.html

moorepants: ok

moorepants:
+1, merge when tests pass.

.travis.yml
```diff
@@ -52,3 +53,5 @@ script:
   - cd docs && make html && cd ..
   # Make sure the library installs.
   - python setup.py install
+  # Run benchmark
+  - python bin/benchmark_pydy_code_gen.py 4 100
```
moorepants: You could make this: `1 10` so it runs faster. Any reason we need this to run with 4 links and 100 time steps?

oliverlee: not really

PR #279, opened 2015-10-17
Fix dependency minimum versions in setup.py

oliverlee:
This change is due to the following commit:
https://github.com/pydy/pydy/commit/9e414bba90fc52a366b7c37977d2247ad34cce51


moorepants:
I thought we agreed to support those that are available in Ubuntu 14.04? Or is this just to fix errors from the theano pr you recently made?

oliverlee:
Yes this is to fix the recent PR. We'll need to support other versions with
another PR.

Issue #206, opened 2015-06-08
The GUI rerun should support changing the initial conditions and time array.


moorepants:
There is no way to change the initial conditions or time array from the GUI. These are important items to be able manipulate.

oliverlee:
Keep in mind that the system may not be defined and just the trajectory is available, so these options won't always be useable.

moorepants:
Yes, these would only be changeable if you are using the `System` class to generate a scene, much like the constants are used now. We need to make sure that if arbitrary trajectories are passed in it is aware.

oliverlee:
The time parameters PR hasn't been merged.

PR #287, opened 2015-10-19
Create text widgets for System time parameters

oliverlee:
This pull request resolves https://github.com/pydy/pydy/issues/206.

Maybe we should move the time parameter widgets so they are horizontally stacked instead of vertically? I don't know how to do that so someone else can, ideally in another pull request.

We also need to have some sort of widget or slider for fps/speedup since it's kind of dumb to tie the simulation time step to the speed of the visualization.

### DIFF
--- a/pydy/viz/scene.py
+++ b/pydy/viz/scene.py
@@ -416,14 +416,17 @@ class Scene(object):
             times = self.system.times
 
         if times is None:
+            tf = (num_time_steps - 1)/self.frames_per_second
             self._scene_info["timeDelta"] = 1.0 / self.frames_per_second
             self._scene_info["startTime"] = 0.0
+            self._scene_info["endTime"] = tf
         else:
             # Assume that times is evenly spaced and monotonic.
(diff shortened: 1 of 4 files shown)


oliverlee:
Screenshots when using a System object:

screen shot 2015-10-20 at 01 46 35

not using a System object:

screen shot 2015-10-20 at 01 47 28

moorepants:
I think you can use container widgets to make columns so that you can arrange things better.

moorepants:
This needs to be updated with master.

moorepants:
Switching to #310

moorepants:
I just tried this out and noticed two things:

1. The animation is very slow. You noted that we need a speed up/slow down knob. I agree.
2. It doesn't seem possible to drag the slider to control the animation anymore.

moorepants:
What's the status with this? I see you added some commits after my comments. Did you address them?

oliverlee:
I noticed the same issues and I haven't looked into it yet.

Issue #290, opened 2015-10-20
code generation fails with large input


oliverlee:
Calls to ```CythonMatrixGenerator.generate()``` will sometimes fail if the input is large enough.

From the [Python documentation](https://docs.python.org/3/library/subprocess.html#subprocess.call):

```
Note: Do not use stdout=PIPE or stderr=PIPE with this function. The child process will block if
it generates enough output to a pipe to fill up the OS pipe buffer as the pipes are not being read
from.
```

We mistakenly do this:
https://github.com/pydy/pydy/blob/b5edaf401df648fa889de9e6e292148f78ae3619/pydy/codegen/cython_code.py#L216-L217

PR #291, opened 2015-10-20
Don't use stdout=PIPE when generating Python modules

oliverlee:
When generating Python modules, use of stdout=PIPE or stderr=PIPE in
subprocess.call() can generate enough output that the OS pipe buffer
blocks, resulting in a compilation failure of generated code. For more
information, see Python documentation here:
https://docs.python.org/3/library/subprocess.html#subprocess.call

This commit resolves https://github.com/pydy/pydy/issues/290.

### DIFF
--- a/pydy/codegen/cython_code.py
+++ b/pydy/codegen/cython_code.py
@@ -213,8 +213,15 @@ setup(name="{prefix}",
             self.write()
             cmd = [sys.executable, self.prefix + '_setup.py', 'build_ext',
                    '--inplace']
-            subprocess.call(cmd, stderr=subprocess.STDOUT,
-                            stdout=subprocess.PIPE)
+            try:
+                DEVNULL = subprocess.DEVNULL
+            except AttributeError:
+                DEVNULL = open(os.devnull, 'wb')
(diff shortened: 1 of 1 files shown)


moorepants:
So, there are no adverse affects? Is all of the codegen output still hidden from the user's view?

oliverlee:
Was that the point originally? All the output is shown now.

moorepants:
This call mimics the autowrap feature in sympy (eventually we should be able to use that machinery but we need some better support). The idea is that this stuff works like at JIT compiler and the user doesn't see anything happening in the background if they don't want to. It should hide the compilation stuff from the user.

oliverlee:
Okay it doesn't show anything now.

moorepants:
This kinds of things really need to be checked on Windows. I can try this weekend, but I don't have time now.

oliverlee:
Great thanks!

moorepants:
We should figure out how to get appveyor running for windows testing and we could also enable a travis mac vm.

moorepants:
As long as this behaves the same as before, you can merge.

oliverlee:
Did you have a chance to try this on Windows?

moorepants:
No, I don't have time till weekend. I'm ok with merging and opening an issue to check it out.

PR #292, opened 2015-10-20
Fix codegen line break

### LINKED ISSUE #263: C code generation line break issue
Hello, while using CythonMatrixGenerator and CMatrixGenerator I stumbled upon the following issue:
the generated C code appears to have some undesirable line breaks which render the code uncompilable. For example, this generated statement

double pydy_2382 = (1.0L/4.0L)pydy_1930pydy_203pydy_204pydy_2339pydy_2
95pydy_345*pydy_60;

yields the following gcc error:

error: expected ‘,’ or ‘;’ before numeric constant
95pydy_345pydy_60;

Of course there shouldn't be a line break in the middle of the variable name pydy_295. That happens when attempting to compile the C code on my own. When I try using the CythonMatrixGenerator.compile() method, the Exception('Failed to compile and import Cython module.') is raised, I guess because of the same issue.
For now I managed to correct the generated c code manually since there are only 2 occurrences of this error, but maybe this is an issue that could be easily fixed in pydy.

Thanks in advance,

Alejandro

oliverlee:
Some generated expressions, particularly those that contain only '\*' and
'/' operands, have no whitespace. In this case, the textwrap module may
break the line in the middle of a symbol or mathematical expression if
the line is long, resulting in code that cannot compile. This commit
adds whitespace before and after the '\*', '/' operands so that textwrap
will have more options for inserting line breaks. The textwrap option
'break_long_words' is also set to True, allowing long lines in the event
that no whitespace is present.

A test has been added that reproduces the bug.

This pull request resolves #263.

This pull request depends on https://github.com/pydy/pydy/pull/291. Please merge that request first.

### DIFF
--- a/pydy/codegen/tests/test_c_code.py
+++ b/pydy/codegen/tests/test_c_code.py
@@ -191,11 +191,11 @@ input_3[6] : [f0(t), f1(t), f2(t), f3(t), f4(t), f5(t)]\
     double pydy_5 = input_3[3];
     double pydy_6 = input_3[4];
     double pydy_7 = input_3[5];
-    double pydy_8 = input_0[6]*input_0[15];
-    double pydy_9 = input_0[6]*input_0[16];
-    double pydy_10 = input_0[6]*input_0[17];
-    double pydy_11 = input_0[6]*input_0[18];
-    double pydy_12 = input_0[6]*input_0[14] + pydy_10 + pydy_11 + pydy_4 +
+    double pydy_8 = input_0[6] * input_0[15];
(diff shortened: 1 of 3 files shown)


moorepants:
This looks good to me.

oliverlee:
I've removed the commit that's included in https://github.com/pydy/pydy/pull/291. Feel free to merge after you've taken a look, assuming there are no other problems.

moorepants:
Looks good. After the #291 merge and travis tests pass, then this is good to go in. Thanks for the fix!

Issue #110, opened 2015-01-20
Consider using PyThreeJS


moorepants:
See this for some examples: http://nbviewer.ipython.org/github/jasongrout/pythreejs/blob/master/examples/Examples.ipynb

We've written some similar code to generate js files to be consumed by threejs. This package is a better approach and would give us IPython interactive capabilities by default.

chrisdembia:
@aymanhab

oliverlee:
They do not have ```TrackballControls```. They have ```OrbitControls```, but this only has 2 rotational degrees of freedom.

I've started a branch with work on this here: https://github.com/oliverlee/pydy/tree/viz-use-pythreejs

Here's a screenshot:
<img width="842" alt="screen shot 2015-10-22 at 18 44 12" src="https://cloud.githubusercontent.com/assets/2965031/10671798/eb920e72-78ec-11e5-990a-93aa2db99a0b.png">


Some notes:
 * Animation is poor. I'm doing it through Python instead of only in JS but I don't know how to hook into the render loop with pythreejs (I didn't see any examples!).
 * Transforms aren't exactly correct. Maybe the geometries in pythreejs need different orientations.

Not everything is implemented (multiple cameras/lights,  parameter editor, etc.)

I'd appreciate any comments or suggestions or fixes.

moorepants:
Sweet! You should contact Jason Grout (the author of the package). I think they use this in SageMathCloud and do fully rotations with 3D objects. If you connect with them they may lead you riht to what you need.

oliverlee:
Okay. The project doesn't look to be active.

moorepants:
I'm pretty sure that sagemathcloud use it, but may be wrong.

Doing the animation through python will have speed issues.

Another thing to check into is Bokeh and whether they will support 3D things. Another option is that VizPy can export to WebGL (at least in the dev version). But this may not solve our problem.

Before you go deep, talk with Jason Grout.

oliverlee:
Probably won't look at it for a while.

I opened some issues on the pythreejs repo.

aymanhab:
@moorepants We'r experimenting with using threejs for visualization in OpenSim and I was wondering if you or anybody on this thread knows the reason for slow animations in PyThreeJS so we can know where the "bodies" are. Thanks.

moorepants:
[...]

PR #297, opened 2015-10-22
WIP Display scenes using pythreejs

oliverlee:
This may resolve https://github.com/pydy/pydy/issues/110


moorepants:
@oliverlee Can you check the :"allow maintainers to merge"? I'm going to get this PR merged this summer so I can use it when I teach dynamics in the fall. I fixed some issues because the dependencies have changed but still don't quite have it working.

PR #280, opened 2015-10-17
Add Python 3.5 to Travis test matrix

oliverlee:

PR #305, opened 2015-12-26
Test against Python 3.5 on Travis

moorepants:
This is the same as PR #280.


moorepants:
@oliverlee This PR looks good now that I have all the packages built. Merging.

PR #282, opened 2015-10-17
Drop minimum dependency versions

oliverlee:

PR #306, opened 2015-12-26
Set minimum dependency versions to match Ubuntu Trusty 14.04 LTS

moorepants:
This is being used in place of PR #282.


moorepants:
This one is good to go now too.

PR #298, opened 2015-10-22
Replace sympy.phyics.mechanics deprecated methods

oliverlee:
This commit resolves https://github.com/pydy/pydy/issues/296.

PR #309, opened 2015-12-26
Replace sympy.phyics.mechanics deprecated methods

moorepants:
Building on #298.


moorepants:
@oliverlee I don't think we can merge this as is. `camera.py` and `light.py` probably don't have good unit tests and when these are run with SymPy 0.7.4.1, which doesn't have these new attributes, the code will fail.

moorepants:
Actually maybe it does work because of this line in pre-0.7.7: https://github.com/sympy/sympy/blob/sympy-0.7.6.1/sympy/physics/mechanics/particle.py#L61

moorepants:
This is good to merge once the tests pass.

PR #317, opened 2015-12-30
removed random choices from time_rhs

yashu-seth:
consistent time measurements cannot be made with random choices

### DIFF
--- a/bin/time_rhs.py
+++ b/bin/time_rhs.py
@@ -29,15 +29,14 @@ constants = list(sm.ordered(sys.constants_symbols))
 
 specifieds = list(sm.ordered(sys.specifieds_symbols))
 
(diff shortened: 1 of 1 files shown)


bin/time_rhs.py
```diff
@@ -37,14 +36,15 @@
 p_dct = dict(zip(constants, p_array))
 
 p = {}
-p[None] = choice([p_array, p_dct])
+# for p[None] 'array' argument type has been assumed
+p[None] = p_array
```
moorepants: Maybe just don't do the runs for None at all since it is the same thing as the line below it.

yashu-seth:
@moorepants I have made the suggested change but I cannot understand why travis is failing for python 2.7.

yashu-seth:
@moorepants Is everything fine now?

PR #322, opened 2016-01-11
Adds in more type checks for System.times.

### LINKED ISSUE #321: Add more checks for times in System

yashu-seth:
Fixes #321.

  commit message.
  in the unit test.)
  follow deprecation cycles.)

### DIFF
--- a/pydy/system.py
+++ b/pydy/system.py
@@ -53,6 +53,7 @@ you must call ``generate_ode_function`` on your own::
 import warnings
 from itertools import repeat
 
+import numpy as np
 import sympy as sm
 from sympy.physics.mechanics import dynamicsymbols
 from scipy.integrate import odeint
@@ -346,11 +347,11 @@ class System(object):
 
     @times.setter
     def times(self, new_times):
-        self._check_times(new_times)
-        self._times = new_times
+        self._times = np.asarray(new_times)
+        self._check_times(self._times)
 
     def _check_times(self, times):
-        if len(times) == 0:
+        if len(times.shape) == 0:
             raise TypeError("Times supplied should be in an array_like format.")
 
         if not all(time >= 0 for time in times):
@@ -358,6 +359,7 @@ class System(object):
 
         if any(time != sort_time for time, sort_time in zip(times, sorted(times))):
             raise ValueError("Times supplied should be in an ascending order.")
+
         return True
 
     @property
--- a/pydy/system.py
+++ b/pydy/system.py
@@ -353,12 +353,11 @@ class System(object):
         if len(times) == 0:
             raise TypeError("Times supplied should be in an array_like format.")
 
-        if not all(time>=0 for time in times):
+        if not all(time >= 0 for time in times):
             raise ValueError("Times supplied must have positive values.")
 
-        if any (time!=sort_time for time,sort_time in zip(times,sorted(times))):
+        if any(time != sort_time for time, sort_time in zip(times, sorted(times))):
             raise ValueError("Times supplied should be in an ascending order.")
-        
         return True
 
     @property
--- a/pydy/system.py
+++ b/pydy/system.py
@@ -354,10 +354,10 @@ class System(object):
         if len(times.shape) == 0:
             raise TypeError("Times supplied should be in an array_like format.")
 
-        if not all(time >= 0 for time in times):
+        if not np.all(times >= 0):
             raise ValueError("Times supplied must have positive values.")
 
-        if any(time != sort_time for time, sort_time in zip(times, sorted(times))):
+        if not np.all(np.diff(times) >= 0):
             raise ValueError("Times supplied should be in an ascending order.")
 
         return True
--- a/pydy/system.py
+++ b/pydy/system.py
@@ -340,22 +340,25 @@ class System(object):
         equations of motion are integrated, numerically.
 
         The object should be in a format which the integration module to be
-        used can accept. Since this attribute is not checked for
-        compatibility, the user becomes responsible to supply it correctly.
+        used can accept.
         """
         return self._times
 
     @times.setter
     def times(self, new_times):
+        self._check_times(new_times)
         self._times = new_times
 
     def _check_times(self, times):
-        """
-        Very basic checking.
-        TODO: add more checking
-        """
         if len(times) == 0:
             raise TypeError("Times supplied should be in an array_like format.")
+
+        if not all(time>=0 for time in times):
+            raise ValueError("Times supplied must have positive values.")
+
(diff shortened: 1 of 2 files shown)


yashu-seth:
@moorepants

pydy/system.py
```diff
@@ -340,22 +340,24 @@ def times(self):
...
 
     def _check_times(self, times):
-        """
-        Very basic checking.
-        TODO: add more checking
-        """
         if len(times) == 0:
             raise TypeError("Times supplied should be in an array_like format.")
+
+        if not all(time >= 0 for time in times):
```
moorepants: `times` should be converted to a numpy array immediately and then these checks can avoid loops.

pydy/tests/test_system.py
```diff
@@ -7,16 +7,18 @@
 import sympy as sm
 from sympy.physics.mechanics import dynamicsymbols
 from scipy.integrate import odeint
-theano = sm.external.import_module('theano')
 
 from ..system import System
 from ..models import multi_mass_spring_damper, n_link_pendulum_on_cart
 from ..utils import sympy_equal_to_or_newer_than, PyDyImportWarning
 
+theano = sm.external.import_module('theano')
```
moorepants: Why move this? Seems like it was in a logical place already, i.e. grouped with the external imports.

yashu-seth: pep8 was showing error may be because there were some more import statements after this.

pydy/system.py
yashu-seth: okay. let me change it.

pydy/tests/test_system.py
moorepants: I see. Well, this is a nonstandard import because we use the custom function. PEP8 are guidelines, not strict rules. In this case I think you should leave it as it was.

yashu-seth: okay.

pydy/system.py
```diff
@@ -340,22 +341,26 @@ def times(self):
...
 
     def _check_times(self, times):
-        """
-        Very basic checking.
-        TODO: add more checking
-        """
         if len(times) == 0:
             raise TypeError("Times supplied should be in an array_like format.")
+
+        times = np.array(times)
```
moorepants: I think asarray is the typical choice for this, see:

http://stackoverflow.com/questions/14415741/numpy-array-vs-asarray

moorepants: Also, this needs to be in the setter, no just the check. We want the times attribute to always be a numpy array, not a list.

yashu-seth:
@moorepants I have made the suggested changes.

yashu-seth:
@moorepants Please have a look at this.

pydy/system.py
```diff
@@ -340,22 +341,25 @@ def times(self):
...
+        self._times = np.asarray(new_times)
+        self._check_times(self._times)
 
     def _check_times(self, times):
-        """
-        Very basic checking.
-        TODO: add more checking
-        """
-        if len(times) == 0:
+        if len(times.shape) == 0:
```
moorepants: This just checks to make sure the array has at least 1 dimension. This doesn't tell you if this is an array or not. The previous check made sure that the length of the array was greater than zero.

`asarray` will throw an error if what you pass in isn't array like. So this check doesn't make any sense. What are you trying to check?

pydy/system.py
```diff
@@ -340,22 +341,25 @@ def times(self):
...
             raise TypeError("Times supplied should be in an array_like format.")
+
+        if not all(time >= 0 for time in times):
```
moorepants: Each value of np.diff(self.times) should be positive.

pydy/system.py
```diff
@@ -340,22 +341,25 @@ def times(self):
...
+            raise ValueError("Times supplied must have positive values.")
+
+        if any(time != sort_time for time, sort_time in zip(times, sorted(times))):
+            raise ValueError("Times supplied should be in an ascending order.")
```
moorepants: Each value of np.diff(self.times) should be positive.

pydy/system.py
```diff
@@ -340,22 +341,25 @@ def times(self):
...
-        """
-        Very basic checking.
-        TODO: add more checking
-        """
-        if len(times) == 0:
+        if len(times.shape) == 0:
             raise TypeError("Times supplied should be in an array_like format.")
+
+        if not all(time >= 0 for time in times):
+            raise ValueError("Times supplied must have positive values.")
```
moorepants: np.all(self.times > 0)

pydy/system.py
```diff
@@ -340,22 +341,25 @@ def times(self):
...
         """
         return self._times
 
     @times.setter
     def times(self, new_times):
-        self._times = new_times
+        self._times = np.asarray(new_times)
+        self._check_times(self._times)
 
     def _check_times(self, times):
```
moorepants: Why is it necessary to pass in times here? Why not just use self.times, or better yet why have this separate function at all? You can just move the if statements into the setter.

pydy/tests/test_system.py
```diff
@@ -260,6 +262,24 @@ def test_specifieds(self):
...
         sys.integrate()
 
+    def test_times(self):
+        times1 = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+        times2 = [0, -2, 7, 3, -5]
+        times3 = [1, 2, 7, 4, 5]
+        times4 = 4
+
+        sys = System(self.kane, times=times1)
+        assert sys.times == times1
```
moorepants: This should use np.testing.assert_allclose.

pydy/system.py
yashu-seth: @moorepants 
```python
>>> x = 4
>>> x = np.asarray(x)
>>> x
array(4)
>>> x.shape
()
>>> len(x.shape)
0
```
```asarray``` does not throw an error for a non array like input. I think it gives an array of zero dimension. Hence checking dimension length gives 0. Please correct me if I am wrong,

pydy/system.py
yashu-seth: Should I remove ```_check_times``` and add all the checks in the setter itself?

yashu-seth: @moorepants I'll remove _check_times and put all the checks in the setter.

yashu-seth: @moorepants ```_check_times``` has been added because ```times``` is checked before every integerate call.

moorepants: Ok.

pydy/system.py
moorepants: Ok, we'll go with this but I find it confusing that you are checking for an "array_like" format with this check.

moorepants:
I'm merging. Please add a note to the release notes for PyDy 0.4.0.

Issue #243, opened 2015-06-18
server.py needs unit tests


moorepants:

karandesai-96:
Well I could start off my contributions by writing unit tests :)

moorepants:
Sounds good. Give it a shot and let us know if you have any questions.

karandesai-96:
>>> server = Server(scene_file=_scene_json_file)
>>> server.run_server()
I can see this is how the server is used in the Python command line.

For testing purposes, a dummy sample of a **_scene_json_file** is required. From where do I generate it ?

PR #325, opened 2016-02-02
Unit tests for pydy/viz/server.py

karandesai-96:
* This PR is in correspondence to issue #243 and contains work on a newly added file under the directory structure as : **pydy/viz/tests/test_server.py**

* These tests are for the file  : **pydy/viz/server.py**

### DIFF
--- a/pydy/viz/tests/test_server.py
+++ b/pydy/viz/tests/test_server.py
@@ -1,5 +1,6 @@
+import io
+import os
 import sys
-import subprocess
 import signal
 
 # For python 2 and python 3 compatibility
@@ -22,38 +23,24 @@ class TestStoppableHttpServer(object):
 
     def test_stoppable_http_server(self):
         self.stoppable_http_server.server_bind()
-        assert self.stoppable_http_server.run
+        assert self.stoppable_http_server.running
 
         self.stoppable_http_server.stop()
-        assert not self.stoppable_http_server.run
+        assert not self.stoppable_http_server.running
 
 
 class TestServer(object):
 
     def __init__(self):
-        self.test_server = Server(directory="../static/",
+        directory = os.path.join(os.path.dirname(__file__), '../static')
+        self.test_server = Server(directory=directory,
                 scene_file="js/tests/sample_data/scene_desc.json")
 
     def test_run_server(self):
-        assert self.test_server.directory == "../static/"
-        assert self.test_server.port == 8000
+        self.test_server.run_server(headless=True)
+        assert self.test_server.httpd.running
 
-        proc = subprocess.Popen(self.test_server.run_server(),
-                                stdin=subprocess.PIPE,
-                                stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE)
-
-        (stdout, stderr) = proc.communicate()
-        sa = self.test_server.httpd.socket.getsockname()
-        assert sa[0] == "127.0.0.1"
-        assert self.test_server.httpd.run
-
-        proc.communicate(signal.SIGINT, timeout=1)
-        (stdout, stderr) = proc.communicate()
-
-        proc.communicate("y", timeout=1)
-        (stdout, stderr) = proc.communicate()
-
-        proc.terminate()
-
-        assert not self.test_server.httpd.run
+        os.kill(os.getpid(), signal.SIGINT)
+        sys.stdin = io.StringIO(u'y')
+        assert not self.test_server.httpd.running
+        assert not self.test_server._thread.is_alive()
--- a/pydy/viz/tests/test_server.py
+++ b/pydy/viz/tests/test_server.py
@@ -28,31 +28,23 @@ class TestStoppableHttpServer(object):
         assert not self.stoppable_http_server.run
 
 
-class TestServer(object):
-
-    def __init__(self):
-        self.test_server = Server(
-                scene_file="js/tests/sample_data/scene_desc.json")
-
-    def test_run_server(self):
-        assert self.test_server.directory == "static/"
-        assert self.test_server.port == 8000
-
-        process = subprocess.Popen(self.test_server.run_server(),
-                                   stdin=subprocess.PIPE,
-                                   stdout=subprocess.PIPE)
-
-        final_port = 8000
-        while self.test_server._check_port(final_port):
-            final_port += 1
-
-        sa = self.test_server.httpd.socket.getsockname()
-        assert sa[0] == "127.0.0.1"
-        assert sa[1] == final_port
-        assert self.test_server.httpd.run
-
-        process.send_signal(signal.SIGINT)
-        out = process.communicate()
-        process.communicate("y")
-        process.kill()
-        assert not self.test_server.httpd.run
+# class TestServer(object):
+#
+#     def __init__(self):
+#         self.test_server = Server(directory="../static/",
+#                 scene_file="js/tests/sample_data/scene_desc.json")
+#
+#     def test_run_server(self):
+#         assert self.test_server.directory == "../static/"
+#         assert self.test_server.port == 8000
+#
+#         process = Process(target=self.test_server.run_server())
+#         process.run()
+#
+#         sa = self.test_server.httpd.socket.getsockname()
+#         assert sa[0] == "127.0.0.1"
+#         assert self.test_server.httpd.run
+#
+#         process.terminate()
+#         process.terminate()
+#         assert not self.test_server.httpd.run
--- a/pydy/viz/tests/test_server.py
+++ b/pydy/viz/tests/test_server.py
@@ -1,4 +1,6 @@
 import sys
+import subprocess
+import signal
 
 # For python 2 and python 3 compatibility
 if sys.version_info < (3, 0):
@@ -7,7 +9,7 @@ else:
     from http.server import SimpleHTTPRequestHandler
     raw_input = input
 
-from ..server import StoppableHTTPServer
+from ..server import StoppableHTTPServer, Server
 
 
 class TestStoppableHttpServer(object):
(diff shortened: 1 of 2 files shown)


karandesai-96:
@moorepants I will keep on adding more tests incrementally. Currently, I wrote tests for the StoppableHTTPServer class in **server.py**.Please view the commit message and the code, and suggest me improvements as I am new to test driven development procedure.

pydy/viz/tests/test_server.py
```diff
@@ -0,0 +1,22 @@
...
+        self.request_handler = SimpleHTTPRequestHandler
+        self.http_server = HTTPServer
+
+        self.stoppable_http_server = StoppableHTTPServer(("127.0.0.1", 8080),
+                                                         self.request_handler,
+                                                         bind_and_activate=False)
+
+    def test_stoppable_http_server(self):
+        self.stoppable_http_server.server_bind()
+        assert self.stoppable_http_server.run is True
```
oliverlee: Avoid comparing boolean values to True or False using == or is.
https://www.python.org/dev/peps/pep-0008/#programming-recommendations

pydy/viz/tests/test_server.py
```diff
@@ -0,0 +1,22 @@
...
+from BaseHTTPServer import HTTPServer
+
+from ..server import StoppableHTTPServer
+
+
+class TestStoppableHttpServer(object):
+
+    def __init__(self):
+        self.request_handler = SimpleHTTPRequestHandler
+        self.http_server = HTTPServer
```
oliverlee: Why are you defining self.request_handler and self.http_server?

pydy/viz/tests/test_server.py
```diff
@@ -0,0 +1,22 @@
+from SimpleHTTPServer import SimpleHTTPRequestHandler
+from BaseHTTPServer import HTTPServer
```
oliverlee: Please maintain Python 3 compatibility.

pydy/viz/tests/test_server.py
karandesai-96: Ok, I'll fix this.

pydy/viz/tests/test_server.py
karandesai-96: I will add it in the next commit

pydy/viz/tests/test_server.py
karandesai-96: Yes, I just realized they are redundant, I'll remove them.

moorepants:
If you need some sample data for a scene check out the tests for the scene class. It generates some base data.

karandesai-96:
* Well, I was able to generate two JSON files, one containing **scene data**, other containing **visualization data**. Only the JSON file containing scene data is of my interest for server class testing.

* The files are generated through scene tests, so they cannot be generated and deleted in the server tests, as if the scene tests fail by chance, the server tests will also fail -- though generating JSON was not the task of server.

* I am thinking of keeping a sample JSON file permanently under the `static` directory exclusively for server tests, this way the tests will stay independent of each other.

* Is this way appropriate? Please suggest better workarounds if any..

karandesai-96:
About my previous comment, I need not generate JSON files for the purpose, I found them under `static/js/tests/sample_data` directory.

oliverlee:
For future reference, the output of one test should never be dependent on another. It's better to generate data at the beginning of the test or to add test data files.

karandesai-96:
I can't find a suitable way to stop the server. Help needed here. The test here won't pass in this commit.

oliverlee:
You can click on the red x to look at the test failure details. You can then look at the console output for each job to see why your test is failing.

The test fails before the server starts running.

karandesai-96:
* Strange.. I gives as OSError, I didn't face it when I ran nosetests on my machine.

* I will have to change directory of **Server** class from `static/` to `../static/` to overcome this error.

* Also, running the server opens the browser by default, which should be separated out in my opinion, as a person might want to open the browser later. We are anyways printing in the command line, that _visit this **URL** from your browser.._

moorepants:
IPython opens the server by default. I think we were copying that functionality here.

karandesai-96:
Shifting the line webbrowser.open(url) to a separate method does not open the server.

Also I need to perform assertions and further shutdown the server by sending it SIGINT and a "Y" as stdin. But on running the tests, the server starts and nothing can be performed further.

I tried to use subprocess, in the previous commit, but the method _subprocess.communicate_ waits for the process to end.

On manually stopping the server, I get an EOFError, due to killing of the subprocess and this test fails.

What should I do for it?

oliverlee:
You call ```communicate()``` twice, the first with no input. As it has no input, the first call will never terminate.

Call ```communicate()``` once and with input and then use ```wait()``` with a timeout.

karandesai-96:
I will try this and return back with updates.

karandesai-96:
@oliverlee @moorepants Hi, while testing `run_server`, the moment I execute it, the server starts and the test does not move further as it waits for the server to stop. I have tried _multiprocessing.Process_, _multiprocessing.Pool_, _subprocess_, _thread_, _threading_.

None seem to just open the server in the background and let me programatically generate an interrupt and further provide a "y" as stdin, to shutdown the server.

I thought the _thread_ or _threading_ modules will surely help, but the method contains a signal.signal, which is only valid in main thread.

What should I do here now ?

oliverlee:
If you comment out the code, I cannot see how the test fails on Travis.

oliverlee:
The subprocess/Popen module is used to spawn new processes and is intended to replace several older modules and functions, namely ```os.system```. https://docs.python.org/3/library/subprocess.html

We want to run the server in another thread as the current behavior blocks. I've submitted https://github.com/pydy/pydy/pull/326 to move execution into a background thread allowing you to stop it easily.

Here's an example:
```python
    def test_run_server(self):
        self.test_server.run_server(headless=True)
        assert self.test_server.httpd.running

        sys.stdin = io.StringIO('y')
        os.kill(os.getpid(), signal.SIGINT)
        assert not self.test_server.httpd.running
        assert not self.test_server._thread.is_alive()
```

oliverlee:
I don't know what directory you're running the tests in, but it fails on travis:

```
======================================================================
ERROR: pydy.viz.tests.test_server.TestServer.test_run_server
----------------------------------------------------------------------
  File "/home/travis/miniconda/envs/test-env/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/pydy/pydy/pydy/viz/tests/test_server.py", line 41, in test_run_server
    proc = subprocess.Popen(self.test_server.run_server(),
  File "/home/travis/build/pydy/pydy/pydy/viz/server.py", line 81, in run_server
    os.chdir(self.directory)
OSError: [Errno 2] No such file or directory: '../static/'
```

You need to set the directory argument of Server during initialization.

karandesai-96:
@oliverlee That is because after one test, the directory cannot be changed to `static/` again as we get into static/ for the first time. 

This happened on my machine too. I executed os.chdir("..") to get out of the static/ directory after doing the tests.

Does travis test in the same container as appveyor ? It seems strange as I could see one test passing and another failing afterwards sometime ago.

oliverlee:
The directory argument is not set correctly because the working directory on travis is likely different than what you are using. Something like this will set the correct directory regardless of your working directory:
```python
os.path.join(os.path.dirname(__file__), '../static')
```
However, I don't know if the server will modify files in place. Maybe it would be better to create a new directory for each test? @moorepants @sahilshekhawat 

I have no idea what appveyor does but it tests Windows whereas travis tests Linux.

karandesai-96:
Ok, so once your PR is merged, I will pull the code and try to complete these tests.

oliverlee:
You can merge the PR and work on your tests before it gets merged.

karandesai-96:
Strange, all tests pass locally but fail on CI... I am looking onto them now.

karandesai-96:
@oliverlee The tests will pass only if I press the enter key manually while `nosetests` is being executed. This is why the tests passed locally but failed on appveyor and travis, due to absence of manual intervention.

I need to provide an EOF probably.

karandesai-96:
I added a boolean, direct_stop to server for facilitating tests without manual intervention of EOF, as I tried many ways to give an EOF programmatically but none worked. Unicode also doesn't have any character for EOF, else I would have sent it through `sys.stdin`.

The `nosetests` pass on my local machine, but tests on **appveyor** and **travis** are again failing. These tests seem to fail just because of unused import statements. I will clear them up.

karandesai-96:
Ok, so Travis CI builds are passing, but Appveyor are failing. Why so ? I can't see the reason, only show that the process exited with status 1.

But it was a great relief looking at the green! 
@oliverlee How about the change in server.py ?

moorepants:
This is the appveyor line that sow failure:

`pydy.viz.tests.test_server.TestServer.test_run_server ... Command exited with code 1`

Appveyor runs the code on windows. You have to make sure to use cross platform python commands, e.g. not os.system() or other things like that.

karandesai-96:
It isn't actually showing the exact error, I do not have Windows distro on my laptop, only Ubuntu. Though i suspect it is `os.chdir` that might be troubling.

moorepants:
`os.chdir` should work on Windows. There are a few Python commands that have trouble on Windows, for example shutil.rmtree.

karandesai-96:
It was `os.path.dirname(__file__)`. I checked it on stackoverflow. Plus the separatoers for Windows and Ubuntu are different, so I have now let the `os` module decide what to keep.

Let's see whether the builds pass or not.

karandesai-96:
Aw shucks, the Windows build failed again. If you have a Windows distro, can you please point out the error ? Appveyor isn't showing precise info.

Linux builds pass, phew..

karandesai-96:
@moorepants @oliverlee If you have a windows distro, please show me the exact error message. I am scratching my head over it since long and it is the last thing required for this task to be done.

The linux builds have already started passing, only Windows is the issue and Appveyor only says the `Process exited with exit code 1.` and nothing more.

yashu-seth:
@karandesai-96 Have you ensured that the changes are consistent with pep8? I remember I got the same error and got rid of it only after fixing pep8 issues.

moorepants:
The minimal pep8 checks we have only run on the Travis system, not appveyor.

yashu-seth:
Ok. Thanks.

karandesai-96:
If anybody has Python set up on a Windows distro please do run it and let me know the error log. I can't make a guess now, and it is the last ting to do in this PR.

moorepants:
FYI, #326 has been merged. Maybe that will help. @oliverlee said that you can't check this on appveyor though, due to some limitation. Maybe we should only have the test that sends the kill signal run on local machines.

karandesai-96:
Sounds like a plan :smile:

PR #326, opened 2016-02-04
Run visualization server in background thread

oliverlee:
This PR runs the visualization server in a background thread instead of the main thread. An option to skip launching the web browser has also been added for debugging.

This will make testing the server easier as the server can be gracefully stopped.

  commit message.
  in the unit test.)
  follow deprecation cycles.)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,6 +68,7 @@ PyDy has hard dependencies on the following software\ [#]_:
 - NumPy_ >= 1.8.1
 - SciPy_ >= 0.13.3
 - SymPy_ >= 0.7.4.1
+- PyWin32 >= 219 (Windows Only)
 
 .. [#] setuptools >= 8.0 is required if development versions of SymPy are used.
 
--- a/appveyor/install.ps1
+++ b/appveyor/install.ps1
@@ -19,7 +19,7 @@ function InstallCondaPackages ($python_home, $spec) {
 
 function main () {
     UpdateConda $env:PYTHON
-    InstallCondaPackages $env:PYTHON "numpy scipy sympy cython nose coverage"
+    InstallCondaPackages $env:PYTHON "numpy scipy sympy cython pywin32 nose coverage"
 }
 
 main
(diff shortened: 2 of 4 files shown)


oliverlee:
You can test server stop in Windows with something like this:
```python
import io
import os
import sys
import win32api
import win32con
from pydy.viz.server import Server

test_server = Server(directory=<somedir>,
                     scene_file="static/js/tests/sample_data/scene_desc.json")
test_server.run_server(headless=True)
sys.stdin = io.StringIO('y')
win32api.GenerateConsoleCtrlEvent(win32con.CTRL_C_EVENT, 0)
```
However the signal is not handled correctly when I run it with ```nosetests```.

oliverlee:
This is ready for review.

moorepants:
I just tested this on Ubuntu 15.10 and it works fine.

What is this solving though? I'd hate to introduce a dependency for little reason. But this seems ok as it can be install via conda and pip.

You need to update the setup.py dependencies to have pywin32 for Windows. We'll also need to include this for Windows in the conda recipes with a selector.

oliverlee:
There's no way to properly send signals in Windows without the package. I think ipython and some other common packages also depend on it.

moorepants:
So did CRTL-C not work in Windows? I feel like it did when I last used it. I still don't understand the purpose of this change.

oliverlee:
It is so that the server code is run in another thread which allows the
main thread to handle input.

moorepants:
I get that part, but what didn't work before this was introduced?

moorepants:
Merging. Please add a note to the release notes. You can commit that directly to master if you like.

oliverlee:
This is already merged? Next time I'll add a note to the release notes in
the PR.

moorepants:
The release note is a bit annoying because if you have multiple pull requests open that are editing the release notes you are bound to get merge conflicts. SymPy keeps their release notes in their wiki because of this. But scipy has a method (https://github.com/scipy/scipy/tree/master/doc/release) that may help...

oliverlee:
Rebasing off master right before merging always works for me.

Issue #20, opened 2014-04-10
Automated Javascript testing for travis


moorepants:
_Issue by **[tarzzz](https://github.com/tarzzz)** from Tuesday Sep 17, 2013 at 06:17 GMT_
_Originally opened as https://github.com/pydy/pydy-viz/issues/60_

---

Integrate JS testing with travis.

PR #335, opened 2016-03-10
Fixes printing bug in the GUI for initial conditions.

moorepants:
My previous implementation doesn't work as expected so I rewrote this to do a
simple replacement on the latex string that seems to work reliably. This stuff,
of course needs unit tests but we need to figure out how to text the notebook
output first.

  commit message.
  in the unit test.)
  follow deprecation cycles.)

### DIFF
--- a/pydy/viz/scene.py
+++ b/pydy/viz/scene.py
@@ -704,9 +704,13 @@ class Scene(object):
 
             val = self._system.initial_conditions[sym]
 
-            sym_0 = sym.xreplace({dynamicsymbols._t: self._system.times[0]})
-
-            desc = latex(sym_0, mode='inline')
+            # TODO : There should be a better way to do this. It would be nice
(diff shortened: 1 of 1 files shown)


moorepants:
@pydy/pydy-developers Review needed.

moorepants:
Note that the previous implementation was subject to this SymPy bug: https://github.com/sympy/sympy/issues/6938

phi is an mpmath function so it tries to evaluate that particular variable that was used in the chaos pendulum example.

oliverlee:
Please open an issue about testing notebooks.
And add a release note (do we still want to as this since the bug was introduced after the last release?) so I can merge.

moorepants:
I don't think a release note is necessary and there is already an issue for testing the javascript, #20.

We need to simply increase the coverage in the gui code.

PR #336, opened 2016-03-11
added drop down using AccordionWidget

mvnnn:


pydy/viz/scene.py
```diff
@@ -752,16 +752,37 @@ def display_ipython(self):
             self._initial_conditions_text_widgets = OrderedDict()
             self._fill_initial_conditions_widgets()
             title = widgets.HTML('<h4 style="margin-left: 5ex;">Initial Conditions</h4>')
-            self._initial_conditions_container.children = ((title, ) +
-                 tuple(v for v in self._initial_conditions_text_widgets.values()))
+            if len(self._initial_conditions_text_widgets) > 5:
+                itemList = []
+                page1 = []
+                for item in self._initial_conditions_text_widgets:
+                    itemList.append(widgets.Text(description=item.name))
```
mvnnn: we need to add value here. i try to find value using dir(symbol) but i didn't find this.
```
widgets.Text(description=item.name, value = ***)
```

pydy/viz/scene.py
```diff
@@ -752,16 +752,37 @@ def display_ipython(self):
             self._initial_conditions_text_widgets = OrderedDict()
             self._fill_initial_conditions_widgets()
             title = widgets.HTML('<h4 style="margin-left: 5ex;">Initial Conditions</h4>')
-            self._initial_conditions_container.children = ((title, ) +
-                 tuple(v for v in self._initial_conditions_text_widgets.values()))
+            if len(self._initial_conditions_text_widgets) > 5:
```
oliverlee: Don't hardcode numbers here. It will be annoying to find and change if necessary.

pydy/viz/scene.py
```diff
@@ -752,16 +752,37 @@ def display_ipython(self):
...
+                    tuple(v for v in self._initial_conditions_text_widgets.values()))
 
             self._constants_container = widgets.Box()
             self._constants_container.padding = "20px"
             self._constants_text_widgets = OrderedDict()
             self._fill_constants_widgets()
             title = widgets.HTML('<h4 style="margin-left: 5ex;">Constants</h4>')
-            self._constants_container.children = ((title, ) +
-                 tuple(v for v in self._constants_text_widgets.values()))
+            if len(self._constants_text_widgets) > 5:
```
oliverlee: Same here.

oliverlee:
Do you have an example that illustrates this change?

mvnnn:
example --->```examples/chaos_pendulum/chaos_pendulum.py```

pydy/viz/scene.py
mvnnn: @oliverlee i have another idea. we display first five variable and put another variable in dropdpwn.

oliverlee:
It looks like you implement an accordion view but only use a single page which doesn't help with the original issue. Also please reference the issue number in this PR.

screen shot 2016-03-11 at 16 31 13

If you're going to go with accordion view, I think it would look better to always use it and just add pages as necessary.

mvnnn:
@oliverlee in this case scroller is best option. i will try to add scroller using ```iframe```.

moorepants:
The above looks pretty good to me. You just need to make a box for the initial conditions too.

pydy/viz/scene.py
mvnnn: @moorepants how to get symbol value . you have any idea ?

moorepants: The initial condition symbol and value are stored in `self._system.initial_conditions`.

mvnnn:
@moorepants thanks, i will do that.

mvnnn:
screenshot from 2016-03-19 21 48 01

pydy/viz/scene.py
mvnnn: @moorepants How to get ```Constants``` value ?

moorepants:
I think an accordion for each of the initial conditions and the constants is preferable. Some systems have long lists of states too.

mvnnn:
@moorepants How to get ```Constants``` value ?

moorepants:
I don't understand what you are asking.

pydy/viz/scene.py
```diff
@@ -752,35 +752,28 @@ def display_ipython(self):
...
-                _initial_item = widgets.Box(children = itemList)
-                self._initial_conditions_container = widgets.Accordion(children=[_initial_item])
-                self._initial_conditions_container.set_title(0, 'Initial Conditions')
-            else :
-                self._initial_conditions_container.children = ((title, ) +
-                    tuple(v for v in self._initial_conditions_text_widgets.values()))
+            _itemList = []
+            _initial_item = []
+            for item in self._initial_conditions_text_widgets:
+                _itemList.append(widgets.Text(description=item.name))
```
mvnnn: Need to add value property here.

pydy/viz/scene.py
```diff
@@ -752,35 +752,28 @@ def display_ipython(self):
...
-                self._constants_container = widgets.Accordion(children=[_constants_item])
-                self._constants_container.padding = "60px"
-                self._constants_container.set_title(0, 'Constants')
-            else :
-                self._constants_container.children = ((title, ) +
-                    tuple(v for v in self._constants_text_widgets.values()))
+            _itemList = []
+            _constants_item = []
+            for item in self._constants_text_widgets:
+                _itemList.append(widgets.Text(description=item.name))
```
mvnnn: Need to add value property here.

Issue #299, opened 2015-12-17
Shape dimensions should only be int or float


yashu-seth:
```python
>>> from pydy.viz.shapes import Sphere,Cube
>>> c = Cube('abcd')
>>> s = Sphere('xyz')
>>> print(s)
Sphere unnamed color:grey material:default radius:xyz
>>> print(c)
Cube unnamed color:grey material:default length:abcd
```

yashu-seth:
@moorepants I don't think this is the expected beaviour. If you also feel the same, I can send a PR.

moorepants:
We work somewhat on the "garbage in/garbage out" principal instead of having tons of type checks, but if you'd like to submit a PR with the type checks I will accept it. Thanks.

yashu-seth:
@moorepants okay i think i'll pass. Can you please get me started with some beginner friendly issues?

moorepants:
This would be a good one to start with: https://github.com/pydy/pydy/issues/296

You have to run the examples and see which ones are using deprecated code.

yashu-seth:
@moorepants thank you. I'll look into it.

oliverlee:
I think the existing PR fixes this but I can't check for a while.

yashu-seth:
@oliverlee yes, I think you have already sent a PR for this.

oliverlee:
@yashu-seth I just realized I won't have wifi for a while. so if you'd like to work on it, please do. You can take a look at the existing PR for a starting point.

moorepants:
Note that the shapes support sympy expressions as dimensions, so they shouldn't "only be int or float".

I actually don't see any reason to make these specific type checks. Python code typically follows duck typing and I'm fine if we use the garbage in/garbage out philosophy here. You simply need to have clear docstrings that say what should be passed in. We don't have to have tons of type checks.

PR #342, opened 2016-03-20
Validate that shape dimension is either int or float value 

abhishekgarg16:
Fix for bug [#299](https://github.com/pydy/pydy/issues/299)

Dimensions passed as attributes to shapes should either be integer or float values. I have added a function check_valid_input() which checks that the dimensions passed are either integer or float values. If it is not so, it raises 'TypeError' in that case.
  commit message.
  directory)
  follow deprecation cycles.)

### DIFF
--- a/pydy/viz/shapes.py
+++ b/pydy/viz/shapes.py
@@ -251,7 +251,6 @@ class Cube(Shape):
 
     def __init__(self, length, **kwargs):
         check_valid_input([length])
-            #raise TypeError('Dimensions should be integer or float values.')
         super(Cube, self).__init__(**kwargs)
         self.geometry_attrs.append('length')
         self.length = length
--- a/pydy/viz/shapes.py
+++ b/pydy/viz/shapes.py
@@ -205,6 +205,11 @@ class Shape(object):
                                 'must provide a mapping to numerical values.')
         return data_dict
 
+def check_valid_input(dimensions):
+    for dimension in dimensions:
+        if not isinstance(dimension, int) and not isinstance(dimension, float):
+            raise TypeError('Dimensions should be integer or float values.')
+
 
 class Cube(Shape):
     """Instantiates a cube of a given size.
@@ -245,6 +250,8 @@ class Cube(Shape):
     """
 
     def __init__(self, length, **kwargs):
+        check_valid_input([length])
+            #raise TypeError('Dimensions should be integer or float values.')
         super(Cube, self).__init__(**kwargs)
         self.geometry_attrs.append('length')
         self.length = length
@@ -297,6 +304,7 @@ class Cylinder(Shape):
 
     """
     def __init__(self, length, radius, **kwargs):
+        check_valid_input([length, radius])
         super(Cylinder, self).__init__(**kwargs)
         self.geometry_attrs += ['length', 'radius']
         self.length = length
@@ -350,6 +358,7 @@ class Cone(Shape):
 
     """
     def __init__(self, length, radius, **kwargs):
+        check_valid_input([length, radius])
         super(Cone, self).__init__(**kwargs)
         self.geometry_attrs += ['length', 'radius']
         self.length = length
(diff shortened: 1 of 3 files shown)


moorepants:
You'll need to write some units tests that ensure this doesn't regress. The tests should try both correct and incorrect input.

moorepants:
What happens if I do this:

```python
s = Cylinder()
s.radius = 'xyz'
```

moorepants:
You also need to run all of the tests locally before pushing. Note that the Ci builds fail a bunch of tests with your new error.

abhishekgarg16:
@moorepants , but when we create an instance of cylinder shape, we need to pass its length and radius. Why do you want to do it as 
`s = Cylinder()`
`s.radius = 'xyz'`

abhishekgarg16:
The code written by me follows the PEP-8 standards.
Can you please help me figure out why all the tests have failed?

abhishekgarg16:
@moorepants , the tests for the shape module were passing strings as shape dimensions. I have modified the tests to have integers or float values as shape dimensions. I have also added tests for 'TypeError' in case the dimensions passed are not integer or float values. I ran all the tests on my local machine and they all passed.

abhishekgarg16:
The issue raised by you in your second comment is valid. If the attributes of an instance are changed after its creation, 'TypeError' is not raised in that case. I am looking into what can be done to fix this. Any idea, how to solve this problem?

moorepants:
This check should happen in the propert setter, not in the `__init__`.

abhishekgarg16:
@moorepants , I have added property setters in each class to verify the same. I was having problems while trying to do it in the parent class, since the number of dimensions is different for each shape. All the tests have passed on my local machine.

abhishekgarg16:
@moorepants , I have added the property setters in the parent class 'Shape'. Also, i have changed some tests in the test_shapes.py file to adapt to the new code. These tests seem to be failing here because the tests themselves are wrong according to the new fix and need to be changed.

pydy/viz/tests/test_scene.py
```diff
@@ -59,8 +59,7 @@ def setup(self):
         self.origin = origin
         self.viz_frame = VisualizationFrame(ceiling, block, sphere)
         self.viz_frame_sym_shape = VisualizationFrame(ceiling, block,
-                                                      Sphere(radius=mass /
-                                                             10.0))
+                                                      Sphere(radius = 10.0))
```
moorepants: These tests are correct. We allow symbols to be passed in as the attributes for shapes so that you can define them symbolically. Please don't change the current tests unless you have very good reason to.

pydy/viz/tests/test_scene.py
```diff
@@ -250,7 +249,7 @@ def test_generate_scene_dict(self):
         scene._generate_simulation_dict()
         scene._generate_scene_dict()
 
-        assert scene._scene_info['objects'][viz_frame_id]['radius'] == 0.1
+        assert scene._scene_info['objects'][viz_frame_id]['radius'] == 10.0
```
moorepants: Do not change tests.

pydy/viz/tests/test_shapes.py
```diff
@@ -57,30 +57,31 @@ def test_shape():
 def test_shape_geometry_with_expressions():
 
     shape = Shape()
-    shape.length = symbols('l')
+    shape.length = 16.0
```
moorepants: All of the changes in this file need to be reverted. These tests specifically test the symbolic functionality for the shape dimensions.

moorepants:
I'm sorry but all of this is not correct. The shape dimensions can be numerical values or sympy symbolics.

pydy/viz/shapes.py
```diff
@@ -679,7 +718,10 @@ def tube_radius(self):
 
     @tube_radius.setter
     def tube_radius(self, new_tube_radius):
-        self._tube_radius = new_tube_radius
+        if not isinstance(new_tube_radius, int) and not isinstance(new_tube_radius, float):
```
oliverlee: `isinstance()` allows for a tuple of object types as the second argument.

moorepants:
Closing as this seems abandoned by the submitter.

PR #350, opened 2016-05-11
MAINT: Fix npendulum example

nmayorov:
Hi! I experienced problems running [this example](http://nbviewer.jupyter.org/github/pydy/pydy/blob/master/examples/npendulum/n-pendulum-control.ipynb). So I made several changes to make it run on newer versions of numpy:

1. Mention that `slycot` is necessary to use `python-control`.
2. Use integer division for computing slices.
3. Convert `equilibrium_point` to a numerical array for ODE integration (not sure how it worked before).

After these changes I was able to run the example with Python 3.4 / numpy 1.11 and Python 2.7 / numpy 1.9. Didn't get any warnings.


moorepants:
In cell 10 can you switch the arguments (forces, particles) to (particles, forces) to avoid the warning.

Also, please execute the whole notebook. The last cells don't seem to have been executed.

The `%install_ext` function is deprecated in IPython. You can pip install the version_information instead.

moorepants:
I'm getting:

```
/home/moorepants/miniconda3/envs/pydy-examples-dev/lib/python3.5/site-packages/matplotlib/animation.py:782: UserWarning: MovieWriter avconv unavailable
  warnings.warn("MovieWriter %s unavailable" % writer)
```

on line 29. I think that the avconv fork has now been merged back with ffmpeg.

nmayorov:
Hey, Jason. I fixed your first remarks (commit amended). 

As for avconv and ffmpeg, I'm not really familiar with them, but I have them both installed on OS X:

- ffmpeg from https://ffmpeg.org/download.html
- avconv from brew

Should I remove `writer='avconv'` and mention that ffmpeg will be used by default, but a user may change it to what is available on his system?

moorepants:
It looks like I was wrong. Libav is still a fork, but ubuntu supports both now. Most changes from libav have been merged into ffmpeg. So you can leave it be or make it use the default. Either is fine.

nmayorov:
Removed `writer` argument and added two notes: at the beginning and before the cell with `animate_pendulum`. Check once more and probably it's good to be merged.

moorepants:
Ok, looks good. Thanks for the contribution!

PR #353, opened 2016-06-09
[WIP] Created a basis on which to discuss EOM class

jbm950:
This example is going to walk through the simulation of a two mass
spring damper system where the equations of motion are provided for
pydy's System by hand. This initial commit hold the base code from eom
formation to simulation and display of results.

### DIFF
--- a/docs/eom_base.rst
+++ b/docs/eom_base.rst
@@ -62,8 +62,8 @@ input combinations, all three of the equations of motion forms will be used. ::
 
     >>> eom1 = eombase.EOM(coordinates, RHS, speeds=speeds, 
     ...                    output_eqns=out_eqns)
-    >>> eom2 = eombase.EOM(states, f_full, mass_matrix_full=mm_full,
-    ...                     num_coordinates=2)
+    >>> eom2 = eombase.EOM(states, f_full, mass_matrix=mm_full,
+    ...                    num_coordinates=2)
     >>> eom3 = eombase.EOM(states, f, mass_matrix=mm, coordinate_derivatives=G)
 
 The system instance will now be initialized and set up to perform simulation. ::
@@ -86,6 +86,8 @@ time simulation of the states is complete. ::
     >>> sys.output_eqns()
     {"kinetic energy": array([[0.0], [...], ...]),
      PE: array([[5.0], [...], ...])}
+    >>> eom1.output_eqns
+    {"kinetic energy": KE, PE: 0.5*k1*x1**2 + 0.5*k2*(x2-x1)**2}
     >>> eom1.output_eqns_results
     {"kinetic energy": array([[0.0], [...], ...]),
      PE: array([[5.0], [...], ...])}
@@ -148,6 +150,22 @@ Next step is to define the equations of motion in multiple forms:
     >>> G = Matrix([u, v])
     >>> RHS = mm_full.LUsolve(f_full)
 
+`Define bodies and loads`
+Now the reference frames, points and particles will be set up so this
+information can be passed into `eombase.EOM` in the form of a bodies and loads
+iterable. ::
+
+    >>> N = ReferenceFrame('N')
+    >>> A = N.orientnew('A', 'Axis', [theta, N.z])
+    >>> O = Point('O')
+    >>> P = O.locatenew('P', l * A.x)
+    >>> Pa = Particle('Pa', P, m)
+
+Now the bodies and loads iterables need to be initialized. ::
+
+    >>> bodies = [Pa]
+    >>> loads = [(P, g * m * N.x)]
+
 The equations of motion are in the form of a differential algebraic equation
 (DAE) and DAE solvers need to know which of the equations are the algebraic
 expressions. This information is passed into `eombase` as a list specifying
@@ -164,16 +182,35 @@ An iterable containing the states now needs to be created for the solvers. ::
 Now the equations of motion instances can be created using the above mentioned
 equations of motion formats. ::
 
-    >>> eom1 = eombase.EOM(states, RHS, alg_con=alg_con_full)
-    >>> eom2 = eombase.EOM(states, f_full, mass_matrix_full=mm_full,
-    ...                    alg_con=alg_con_full)
+    >>> eom1 = eombase.EOM(states, RHS, alg_con=alg_con_full, bodies=bodies,
+    ...                    loads=loads)
+    >>> eom2 = eombase.EOM(states, f_full, mass_matrix=mm_full,
+    ...                    alg_con=alg_con_full, num_coordinates=2)
     >>> eom3 = eombase.EOM(states, f, mass_matrix=mm, coordinate_derivatives=G,
-    ...                    alg_con=alg_con)
+    ...                    alg_con=alg_con, num_coordinates=2, num_speeds=2)
 
 Lastly here are some attributes accessible from the `EOM` class. ::
 
     >>> eom1.states
     (x, y, u, v, lam)
+    >>> eom2.coordinates
+    (x, y)
+    >>> eom3.speeds
+    (u, v)
+    >>> eom1.rhs
+    Matrix([[u(t)], [v(t)], [(-g*y(t) + u(t)**2 + v(t)**2)*x(t)/l**2], [(-g*y(t) + u(t)**2 + v(t)**2)*y(t)/l**2], [m*(-g*y(t) + u(t)**2 + v(t)**2)/l**2]])
+    >>> eom2.forcing_full
+    Matrix([u, v, 0, 0, u**2 + v**2 - g*y])
+    >>> eom2.mass_matrix_full
+    Matrix([[1, 0, 0, 0, 0], [0, 1, 0, 0, 0], [0, 0, 1, 0, -x/m], [0, 0, 0, 1, -y/m], [0, 0, 0, 0, l**2/m]])
+    >>> eom3.forcing
+    Matrix([0, 0, u**2 + v**2 - g*y])
+    >>> eom3.mass_matrix
+    Matrix([[1, 0, -x/m], [0, 1, -y/m], [0, 0, l**2/m]])
+    >>> eom1.dynamic_symbols()
+    (x, y, u, v, lam)
+    >>> eom1.constant_symbols()
+    (m, l, g)
 
 ========================================
 Simple Pendulum Theta Coordinate Example
@@ -230,7 +267,7 @@ The `LagrangesMethod` instance can pass an instance of eombase using its
 eombase rather than making the user pass everything in manually. For instance
 it will automatically change the equations to first order form. ::
 
-    >>> EOM = l.to_eom()
+    >>> eom = l.to_eom()
     >>> sys = System(EOM)
 
 Now that the system is set up, a simple time simulation will be performed. ::
@@ -252,7 +289,7 @@ step. ::
 
 Here are some additional attributes accessible from the `eombase.EOM` class. ::
 
-    >>> EOM.bodies
+    >>> eom.bodies
     [Pa]
-    >>> EOM.loads
+    >>> eom.loads
     [(P, g * m * N.x)]
--- a/docs/eom_base.rst
+++ b/docs/eom_base.rst
@@ -217,26 +217,18 @@ equations of motion formats. ::
     >>> symsystem2 = system.SymbolicSystem(states, comb_implicit_rhs, 
     ...                                    mass_matrix=comb_implicit_mat,
     ...                                    alg_con=alg_con_full,
-    ...                                    num_coordinates=2)
+    ...                                    coord_idxs=(0, 1))
     >>> symsystem3 = system.SymbolicSystem(states, dyn_implicit_rhs, 
     ...                                    mass_matrix=dyn_implicit_mat,
     ...                                    coordinate_derivatives=kin_explicit_rhs,
-    ...                                    alg_con=alg_con, num_coordinates=2, 
-    ...                                    num_speeds=2)
+    ...                                    alg_con=alg_con, coord_idxs=(0, 1), 
+    ...                                    speed_idxs=(2, 3))
 
-The `SymbolicSystem` class determines which of the states are considered
-coordinates by knowing the number of coordinates and assuming that the states
-were input in this order: (coordinates, speeds, other). This means if
-num_speeds is specified and num_coordinates is not, the class will not know
-which of the states to consider speeds and will cause an error This also means
-that if just the states are passed in, the class will not be able to determine
-which of the states are coordinates or speeds. ::
+The `SymbolicSystem` class can determine which of the states are considered
+coordinates or speeds by passing in the indexes of the coordinates and speeds.
+If these indexes are not passed in the object will not be able to differentiate
+between coordinates and speeds. ::
 
-    >>> symsystem4 = system.SymbolicSystem(states, comb_implicit_rhs, 
-    ...                                    mass_matrix=comb_implicit_mat,
-    ...                                    num_speeds=2)
-    SyntaxError: Need to specify the number of coordinates if specifying the
-                 number of speeds
     >>> symsystem1.coordinates
     AttributeError: The coordinates were not specified
     >>> symsystem2.speeds
--- a/docs/eom_base.rst
+++ b/docs/eom_base.rst
@@ -41,7 +41,7 @@ Next step is to define the equations of motion in multiple forms:
     ...                  -(k1+k2)*x1 + k2*x2 - (c1+c2)*u1 + c2*u2,
     ...                  k2*x1 - k2*x2 + c2*u1 - c2*u2])
     >>> G = Matrix([u1, u2])
-    >>> RHS = mm_full.inv()*f_full
+    >>> RHS = mm_full.LUsolve(f_full)
 
 Now the various forms of varible entry to `eombase` will be initiated ::
 
@@ -59,16 +59,15 @@ whole system is going to be determined. ::
 The `eombase` instances are now ready to be initialized. To showcase possible
 input combinations, all three of the equations of motion forms will be used. ::
 
-    >>> eom1 = eombase.EOM(coordinates=coordinates, speeds=speeds, rhs=RHS,
+    >>> eom1 = eombase.EOM(coordinates, RHS, speeds=speeds, 
     ...                    output_eqns=out_eqns)
-    >>> eom2 = eombase.EOM(states=states, mass_matrix_full=mm_full,
-    ...                    forcing_full=f_full, num_coordinates=2)
-    >>> eom3 = eombase.EOM(states=states, mass_matrix=mm, forcing=f,
-    ...                    kinematics=G)
+    >>> eom2 = eombase.EOM(states, f_full, mass_matrix_full=mm_full,
+    ...                     num_coordinates=2)
+    >>> eom3 = eombase.EOM(states, f, mass_matrix=mm, coordinate_derivatives=G)
 
 The system instance will now be initialized and set up to perform simulation. ::
 
-    >>> sys = System(eom2)
+    >>> sys = System(eom1)
     >>> sys.times = linspace(0, 10, num=100)
     >>> sys.constants = {m1: 10.0, c1: 10.0, k1: 10.0, m2: 5.0, c2: 5.0, k2:
     ...                  5.0}
@@ -78,15 +77,15 @@ Now the actual system simulation is ready to be run. ::
 
     >>> out = sys.integrate()
     >>> out
-    numpy.array([[1.0, 0.0, 0.0, 0.0],
-                 [...]])
+    array([[1.0, 0.0, 0.0, 0.0], [...], ...])
 
 The user specified output equations should be able to be determine now that the
 time simulation of the states is complete. ::
 
     >>> sys.output_eqns()
+    {"kinetic energy": array([[0.0], [...], ...])}
     >>> eom1.output_eqns_results
-    {"kinetic energy": numpy.array([[0.0], [...], ...])}
+    {"kinetic energy": array([[0.0], [...], ...])}
 
 With the simulation completed the output trajectories can be plotted using
 matplotlib. ::
@@ -139,7 +138,7 @@ Next step is to define the equations of motion in multiple forms:
     ...                   [0, 0, 0, 0, l**2/m]])
     >>> f_full = Matrix([u, v, 0, 0, u**2 + v**2 - g*y])
     >>> G = Matrix([u, v])
-    >>> RHS = mm_full.inv()*f_full
+    >>> RHS = mm_full.LUsolve(f_full)
 
 The equations of motion are in the form of a differential algebraic equation
 (DAE) and DAE solvers need to know which of the equations are the algebraic
@@ -157,11 +156,16 @@ An iterable containing the states now needs to be created for the solvers. ::
 Now the equations of motion instances can be created using the above mentioned
 equations of motion formats. ::
 
-    >>> eom1 = eombase.EOM(states=states, rhs=RHS, alg_con=alg_con_full)
-    >>> eom2 = eombase.EOM(states=states, mass_matrix_full=mm_full,
-    ...                    forcing_full=f_full, alg_con=alg_con_full)
-    >>> eom3 = eombase.EOM(states=states, mass_matrix=mm, forcing=f,
-    ...                    kinematics=G, alg_con=alg_con)
+    >>> eom1 = eombase.EOM(states, RHS, alg_con=alg_con_full)
+    >>> eom2 = eombase.EOM(states, f_full, mass_matrix_full=mm_full,
+    ...                    alg_con=alg_con_full)
+    >>> eom3 = eombase.EOM(states, f, mass_matrix=mm, coordinate_derivatives=G,
+    ...                    alg_con=alg_con)
+
+Lastly here are some attributes accessible from the `EOM` class. ::
+
+    >>> eom1.states
+    (x, y, u, v, lam)
 
 ========================================
 Simple Pendulum Theta Coordinate Example
@@ -215,19 +219,32 @@ Now the lagrangian and force list can be created and with these an instance of
 
 The `LagrangesMethod` instance can pass an instance of eombase using its
 `.to_eom()` method. This allows the class to handle all of the formatting for
-eombase rather than making the user pass everything in manually. ::
+eombase rather than making the user pass everything in manually. For instance
+it will automatically change the equations to first order form. ::
 
-    >>> sys = System(l.to_eom())
+    >>> EOM = l.to_eom()
+    >>> sys = System(EOM)
 
 Now that the system is set up, a simple time simulation will be performed. ::
 
     >>> sys.times = linspace(0, 10, num=100)
     >>> sys.constants = {m: 10, l: 5, g: 9.8}
     >>> sys.initial_conditions = {theta: 60, thetad: 0}
-    >>> out = sys.integrate()
+    >>> sys.integrate()
+    array([[60.0, 0.0], [...], ...])
 
 Display the kinetic energy change in time (obtained from the particle in the
 bodies list). The kinetic energies are displayed in the order listed in the
-`bodies` list ::
+`bodies` list. The last column is the kinetic energy of the whole system and is
+simply the addition of all the other kinetic energies in the array at each time
+step. ::
 
-    >>> KE = sys.body_kinetic_energies()
+    >>> sys.body_kinetic_energies()
+    array([[0.0, 0.0], [...], ...])
+
+Here are some additional attributes accessible from the `eombase.EOM` class. ::
+
+    >>> EOM.bodies
+    [Pa]
+    >>> EOM.loads
+    [(P, g * m * N.x)]
--- a/docs/eom_base.rst
+++ b/docs/eom_base.rst
@@ -84,9 +84,11 @@ The user specified output equations should be able to be determine now that the
 time simulation of the states is complete. ::
 
     >>> sys.output_eqns()
-    {"kinetic energy": array([[0.0], [...], ...])}
+    {"kinetic energy": array([[0.0], [...], ...]),
+     PE: array([[5.0], [...], ...])}
     >>> eom1.output_eqns_results
-    {"kinetic energy": array([[0.0], [...], ...])}
+    {"kinetic energy": array([[0.0], [...], ...]),
+     PE: array([[5.0], [...], ...])}
 
 With the simulation completed the output trajectories can be plotted using
 matplotlib. ::
--- a/docs/eom_base.rst
+++ b/docs/eom_base.rst
@@ -50,11 +50,12 @@ Now the various forms of varible entry to `eombase` will be initiated ::
     >>> states = (x1, x2, u1, u2)
 
 This example will also show `system`'s ability to calculate an arbitrary user
-defined functions along with the states. Specifically the kinetic energy of the
-whole system is going to be determined. ::
+defined functions along with the states. Specifically the kinetic energy and
+potential energy  of the whole system is going to be determined. ::
 
     >>> KE = 0.5 * (m1*u1**2 + m2*u2**2)
-    >>> out_eqns = {"kinetic energy": KE}
+    >>> PE = symbols("PE")
+    >>> out_eqns = {"kinetic energy": KE, PE: 0.5*k1*x1**2 + 0.5*k2*(x2-x1)**2}
 
 The `eombase` instances are now ready to be initialized. To showcase possible
 input combinations, all three of the equations of motion forms will be used. ::
--- a/docs/eom_base.rst
+++ b/docs/eom_base.rst
@@ -66,6 +66,27 @@ input combinations, all three of the equations of motion forms will be used. ::
     ...                    num_coordinates=2)
     >>> eom3 = eombase.EOM(states, f, mass_matrix=mm, coordinate_derivatives=G)
 
+There are some potential problems that can occur during the initialization of
+the eombase.EOM class. First, the class assumes if it recieves
+coordinate_derivatives that the input equations of motion are in form [3]. This
+means it will produce an error if a mass matrix is not recieved. Similarly if a
+mass matrix is recieved and coordinate_derivatives are not specified, the class
+is assuming that the equations of motion have been put in in form [2]. If the
+equations of motion are input in form [1], the class will not have a notion of
+the mass matrix or the forcing vector and as such will return errors if there
+is an attempt to access these attributes. ::
+
+    >>> eom4 = eombase.EOM(states, f, coordinate_derivatives=G)
+    SyntaxError: Need to specify a mass matrix for eom form [3]
+    >>> eom1.mass_matrix
+    AttributeError: Mass matrix is not specified for eom form [1]
+    >>> eom1.mass_matrix_full
+    AttributeError: Mass matrix full is not specified for eom form [1]
+    >>> eom1.forcing
+    AttributeError: Forcing vector is not specified for eom form [1]
+    >>> eom1.forcing_full
+    AttributeError: Full forcing vector is not specified for eom form [1]
+
 The system instance will now be initialized and set up to perform simulation. ::
 
     >>> sys = System(eom1)
@@ -80,9 +101,13 @@ Now the actual system simulation is ready to be run. ::
     >>> out
     array([[1.0, 0.0, 0.0, 0.0], [...], ...])
 
-The user specified output equations should be able to be determine now that the
-time simulation of the states is complete. ::
+The user specified output equations should be able to be determined now that the
+time simulation of the states is complete. These equations are calculated by
+sys.output_eqns and so if there is an attempt to see the results first an error
+will be returned. ::
 
+    >>> eom.output_eqns_results
+    ValueError: The output equations first need to be calculated
     >>> sys.output_eqns()
     {"kinetic energy": array([[0.0], [...], ...]),
      PE: array([[5.0], [...], ...])}
@@ -150,7 +175,6 @@ Next step is to define the equations of motion in multiple forms:
     >>> G = Matrix([u, v])
     >>> RHS = mm_full.LUsolve(f_full)
 
-`Define bodies and loads`
 Now the reference frames, points and particles will be set up so this
 information can be passed into `eombase.EOM` in the form of a bodies and loads
 iterable. ::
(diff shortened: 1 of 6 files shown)


examples/two_mass_spring_damper/run.py
```diff
@@ -0,0 +1,53 @@
...
+                  k2*x1 - k2*x2 + c2*u1 - c2*u2])
+
+# Form the remaining input lists for the equations of motion class
+
+coordinates = (x1, x2)
+speeds = (u1, u2)
+
+# Initialize the equation of motion class
+
+eom = eombase.EOM(coordinates, speeds, mass_matrix, forcing)
```
moorepants: This should be able to accept multiple forms of the eoms. It may be good to look at the three options available for the ODEFunctionGenerator class in PyDy for examples. We may simply need to add one more option to handle DAEs.

moorepants: It would also be nice to support "output equations" here, i.e. those equations that are algebraically dependent on the state. For example, kinetic energy of the system.

jbm950: I'm confused by your second comment. Are you suggesting that the class `EOM` should be able to produce the kinetic energy and such of the system?

moorepants: I'm suggesting that it may be a good idea to have the EOM class contain additional arbitrary equations that are functions of the states, parameters, etc. One example of an equation such as this is an expression for kinetic energy.

moorepants: The auxiliary equations in KanesMethod are also examples.

jbm950:
Was something like that generally what you had in mind?

examples/two_mass_spring_damper/run.py
```diff
@@ -0,0 +1,86 @@
...
+# Initialize the equation of motion class using the three forms accepted by
+# ODEFunctionGenerator
+#    [1] x' = F(x, t, r, p)
+#
+#    [2] M(x, p) x' = F(x, t, r, p)
+#
+#    [3] M(q, p) u' = F(q, u, t, r, p)
+#        q' = G(q, u, t, r, p)
+
+eom1 = eombase.EOM(coordinates, speeds, rhs=RHS, output_eqns=out_eqns)
```
moorepants: `rhs` can probably be a required arg since all forms of the equations have a right hand side.

examples/two_mass_spring_damper/run.py
```diff
@@ -0,0 +1,86 @@
...
+eom2 = eombase.EOM(coordinates, speeds, mass_matrix_full=mm_full,
+                   forcing_full=f_full)
```
moorepants: One thing that was raised on the other pydy pr about this was whether a general EoM object would need to differentiate between the coordinates and speeds or could simple a state vector be passed in.

moorepants:
> Was something like that generally what you had in mind?

Yes, now we need to think about how to incorporate the DAE forms from Lagrange's (or other) methods. And think about what are the attributes and methods of EoM.

examples/two_mass_spring_damper/run.py
jbm950: Your comment confuses me. `rhs` used in `eom1` is specifically F(x, t, r, p) from form [1]. If it were to be generally the right and side of any of the other two forms then the input would have different meanings in different situations. Also this would get rid of `forcing` and `forcing_full` keyword arguments.

moorepants: All three sets of equations have a "right hand side" as shown, i.e. whatever is on the right hand side of the equals sign. Thus for every form lists, the user must always pass in a right hand side of the equations. But the coefficient matrices on the left hand side are optional.

In the ODEFunctionGenerator class I'm pretty sure it always requires the user to pass in the right hand side of whatever form of the equations they are using.

jbm950: Currently in `KanesMethod` and `LagrangesMethod` the right hand side of [1] is obtained by `*Method.rhs()`, the right hand side of [2] is obtained by `*Method.forcing_full` and the right hand side of [3] is obtained by `*Method.forcing`.

From the docstring of `ODEFunctionGenerator`
```
        right_hand_side : SymPy Matrix, shape(n, 1)
            A column vector containing the symbolic expressions for the
            right hand side of the ordinary differential equations. If the
            right hand side has been solved for symbolically then only F is
            required, see form [1]; if not then the mass matrix must also be
            supplied, see forms [2, 3].
```
I suppose the confusion is in not having a more formal name for the right hand side of [1] than simply "rhs"

moorepants: Yes, the name is too general and confusing.

jbm950: Do you have any ideas on something we can change it to? Maybe something like `first_order_form_rhs` or `fof_rhs` for brevity. This is assuming that [1] is the only one represented first order form as implied by the docstring for `rhs()` in `KanesMethod`

```
    def rhs(self, inv_method=None):
        """ Returns the system's equations of motion in first order form.
```

moorepants: I've been working on this stuff for about 8 years now and have yet to devise a good textual naming for all this stuff. I'm still not even sure I'm using all the "correct" terminology. Some words to throw out there:

- ordinary differential equation
- implicit ode
- explicit ode
- differential algebraic equation
- right hand side
- mass matrix
- coefficient matrix
- first order ode
- second order ode

I not sure there is a consensus on how to describe the different "forms" of these ODE/DAEs. And when you try you end up with very verbose words. Maybe check the math and dynamics literature for some ideas.

jbm950: `Maybe check the math and dynamics literature for some ideas.` 

I'm up for doing the research but I don't know where I would start. Do you have a suggestion?

examples/two_mass_spring_damper/run.py
moorepants: Wikipedia and wolfram.

I've been following kane's book to fix up the kane's notation.

I'm sure that people have variations on the lagrange notations since his stuff is so old.

For general ode stuff look up an ode book, but if you could find a popular text book that was about dynamics that covers this stuff that would be good. Maybe Greenwood?

jbm950: `self.coordinates` is only used in two places in `System`: being combined with `self.speeds` to form `self.states` and in `_args_for_gen_ode_func`. The latter of the two places I've chased down and has currently brought me to the internals of `ode_function_generators.LambdifyODEFunctionGenerator` where I am currently searching to see if the coordinates are actually needed separate from the speeds. I'll post here when I find the answer.

moorepants: It probably isn't needed internally to the ode generators. But if you were to create methods on system like:

```python
sys.plot_coordinate_trajectories()
```

It would be nice to have them separate. You can work with just the state vector for visualization and further work with system but for multbody systems, more often than not, you look at the coordinates and speeds separately because they are different scales. You could even imagine how knowing the units of the states would help improve automatic viz and other things.

moorepants: So I guess we need to decide if System is very general or is System made for multibody systems. That was what some of the discussion in pr #345 was about. I personally would like System to be a multibody system so that it is smart enough to do normal things you want to do with a multibody system. But a `MultibodySystem` could be a subclass to a more general `System`. Systems should also tie nicely into pydy.viz. Right now it is a bit clunky, but a System could be aware of the bodies and loads in the system and make a default visualization.

jbm950: `So I guess we need to decide if System is very general or is System made for multibody systems. `
I would say having a general class that is subclassed would provide a broader base for future functionality.

Having just a single variable `states` would give the class less information overall to work with but would be more compact. If subclasses need to know which states are coordinates and which are speeds they could have a simple input variable for the number of coordinates and assume the first elements that correspond to that number are the coordinates and the rest are speeds. 

`You could even imagine how knowing the units of the states would help improve automatic viz and other things.`

I would think this could be very simply by inputting a list of unit names to either `EOM` or `System` where they positionally correspond to coordinates/speeds in the `states` variable. This solution wouldn't require the coordinates and speeds to be separate variables. Example code
```
eom = EOM(states=[q1, q2, q2, u1, u2, u3], units=['m', 'm', 'mm', 'm/s', 'cm/s', 'mm/s'])
```

moorepants: That all sounds fine about system being general.

For units, we should think about how we may utilize the units functionality already available in sympy.

jbm950:
I've begun adding other examples for the API of `eombase`. One addition is the ability to pass in coordinates and speeds together or a single states iterable.

examples/simple_pendulum/run.py
```diff
@@ -39,3 +40,21 @@
...
+sys.times = linspace(0, 10, num=100)
+sys.constants = {m: 10, l: 5, g: 9.8}
+sys.initial_conditions = {theta: 60, thetad: 0}
+
+# Simulate the system
+out = sys.integrate()
+
+# Display the kinetic energy change in time (obtained from the particle in the
+# bodies list)
+KE = sys.body_kinetic_energies()
```
moorepants: What will KE be? Is this similar to out? i.e. an array of trajectories for kinetic energy?

examples/simple_pendulum/run2.py
```diff
@@ -0,0 +1,55 @@
...
+                  [0, 0, 1, 0, -x/m],
+                  [0, 0, 0, 1, -y/m],
+                  [0, 0, 0, 0, l**2/m]])
+f_full = Matrix([u, v, 0, 0, u**2 + v**2 - g*y])
+
+# Form the rhs of q' = G(q, u, t, r, p). Kinematic equation
+G = Matrix([u, v])
+
+# Form the rhs of the dynamics equations
+RHS = mm_full.inv()*f_full
```
moorepants: Typically best to use `mm_full.LUsolve(f_full)` instead of explicit inversion.

examples/simple_pendulum/run2.py
```diff
@@ -0,0 +1,55 @@
...
+#    [2] M(x, p) x' = F(x, t, r, p)
+#
+#    [3] M(q, p) u' = F(q, u, t, r, p)
+#        q' = G(q, u, t, r, p)
+
+eom1 = eombase.EOM(states=states, rhs=RHS, alg_con=alg_con_full)
+eom2 = eombase.EOM(states=states, mass_matrix_full=mm_full, forcing_full=f_full,
+                   alg_con=alg_con_full)
+eom3 = eombase.EOM(states=states, mass_matrix=mm, forcing=f, kinematics=G,
+                   alg_con=alg_con)
```
moorepants: These look good. Some comments on args versus kwargs. Some of these things should probably be args instead of kwargs. For example, each system requires that there be coordinates and speeds OR states. So:

```python
EOM(coordinates, speeds, ...)
EOM(states, ...)
```

Seems more natural and avoids using the kwargs.

Also, rhs is always required. So

```python
EOM(states, rhs, mass_matrix=M)
```

If you look carefully at the API here:

https://github.com/pydy/pydy/blob/master/pydy/codegen/ode_function_generators.py#L186

You'll see that it has a set of required args and a set of optional args. It tries to minimize the total number of distinct args, e.g. only mass_matrix not mass_matrix and mass_matrix_full. This then relies on a good docstring to show how to pass in the different combinations of args/kwargs. I prefer this approach more than making a ton of only optional kwargs (I'm biased though, as I wrote the codegen api).

moorepants: In fact, is there any reason not to copy the codegen api so that we have consistency throughout the code?

examples/two_mass_spring_damper/run.py
```diff
@@ -0,0 +1,94 @@
...
+                  [0, 0, m1,  0],
+                  [0, 0,  0, m2]])
+f_full = Matrix([u1,
+                 u2,
+                 -(k1+k2)*x1 + k2*x2 - (c1+c2)*u1 + c2*u2,
+                 k2*x1 - k2*x2 + c2*u1 - c2*u2])
+
+# Form the rhs of q' = G(q, u, t, r, p). Kinematic equation
+
+G = Matrix([u1, u2])
```
moorepants: I'm pretty sure that lagrange's method spits out q, q', and q'', which are not in explicit first order form, i.e. q'=u hasn't been used to substitute the u's for the q'. This means the lagrange's method gives the equations in second order form instead of first order form, right? How will this class deal with that?

examples/two_mass_spring_damper/run.py
```diff
@@ -0,0 +1,94 @@
...
+
+coordinates = (x1, x2)
+speeds = (u1, u2)
+states = (x1, x2, u1, u2)
+
+# Form the kinetic energy to attach to one of the EOM instances as an output
+# equation
+
+KE = 0.5 * (m1*u1**2 + m2*u2**2)
+out_eqns = {"kinetic_energy": KE}
```
moorepants: Since everything is a variable in the equations of motion, this should map a symbol to an expression instead of mapping a string to an expression. That seems to be more consistent.

Also, should we allow simply an iterable of expressions too? That's the way most things are passed into the `*Method`s classes, for example.

examples/two_mass_spring_damper/run.py
```diff
@@ -0,0 +1,94 @@
...
+#    [2] M(x, p) x' = F(x, t, r, p)
+#
+#    [3] M(q, p) u' = F(q, u, t, r, p)
+#        q' = G(q, u, t, r, p)
+
+eom1 = eombase.EOM(coordinates=coordinates, speeds=speeds, rhs=RHS,
+                   output_eqns=out_eqns)
+eom2 = eombase.EOM(states=states, mass_matrix_full=mm_full, forcing_full=f_full,
+                   num_coordinates=2)
+eom3 = eombase.EOM(states=states, mass_matrix=mm, forcing=f, kinematics=G)
```
moorepants: How about:

```python
eom1 = eombase.EOM(coordinates, speeds, RHS, output_eqns=out_eqns)
eom2 = eombase.EOM(states, forcing_full, mass_matrix=mm_full, num_coordinates=2)
eom3 = eombase.EOM(states, f, mass_matrix=mm, kinematics=G)
```

Why is num_coordinates needed? And how do you know what order the coordinates are in wrt to the state vector?

examples/two_mass_spring_damper/run.py
```diff
@@ -0,0 +1,94 @@
...
+# Simulate the system
+
+out = sys.integrate()
+
+
+# Display the kinetic energy from the first EOM class
+# Returns a dictionary whith the same keys as output_eqns but the values are the
+# numerical results of the simulation
+
+print(eom1.output_eqns_results)
```
moorepants: What does this print?

examples/two_mass_spring_damper/run.py
moorepants: I think we need a way to pass in the bodies and loads too.

Also, can this work:

```python
EOM(kanes_method_instance)
EOM(lagranges_method_instance)
```
?

Or should this be how you do it:

```
kane = KanesMethod(...)
eom_instance = kane.to_eom()
```
?

examples/two_mass_spring_damper/run.py
moorepants: This made me think about why we aren't using documentation, instead of just the input file, to show things. For example you could do:

```python
>>> kane = KanesMethod(...)
>>> kane.states
Matrix([q1, q2, u1, u2])
>>> eom = kane.to_eom()
>>> eom.states
Matrix([q1, q2, u1, u2])
>>> sys = System(eom)
>>> sys.integrate()
array([1.0, .2.0, 3.0, 4.0])
>>> sys.output_equation_trajectories()
array([5.0, 4.0, 6.0])
```

This shows the desired input and output, is more akin to a unit test, and is called "documentation driven development" if you write this first (as you are doing).

examples/simple_pendulum/run.py
jbm950: I think that would be a good idea to have the same output. sys.integrate() seems to give the trajectories such that each column represents a single state. I feel this could be put to the same use here where each column represents a single body.

moorepants: Same style of output sounds good. But what are "kinetic energies"? Each body/particle has kinetic energy and then the whole system has kinetic energy, as does any combination of bodies and particles. This needs clearer definition.

examples/simple_pendulum/run2.py
jbm950: `"For example, each system requires that there be coordinates and speeds OR states."`
I was thinking about allowing the user to pass in either of the two in which case I think kargs would make more sense. In the very least coordinates/states could be a single argument and speeds could be a karg. The class could then test input format by seeing if speeds was passed. By making them all arg inputs could cause errors when using a just states. Take the code snippet you posted later in the changes for example.

```
eom1 = eombase.EOM(coordinates, speeds, RHS, output_eqns=out_eqns)
eom2 = eombase.EOM(states, forcing_full, mass_matrix=mm_full, num_coordinates=2)
eom3 = eombase.EOM(states, f, mass_matrix=mm, kinematics=G)
```

In eom2, forcing_full would be passed to the speeds variable if the class was set up for `__init__(coordinates/states, speeds, forcing/rhs, ...)` If it is decided to only use a single mode of input I believe using `states` would be the better option.

(writing response to second question now)

moorepants: You can use `__init__(*args, **kwargs)` and parse the input in any fashion you want to.

moorepants: We do this in the VisualizationFrame class (probably too much).

You could do this:

```python
eom1 = eombase.EOM(coordinates, RHS, speeds=speeds, output_eqns=out_eqns)
eom2 = eombase.EOM(states, forcing_full, mass_matrix=mm_full, num_coordinates=2)
eom3 = eombase.EOM(states, f, mass_matrix=mm, kinematics=G)
```

jbm950: `In fact, is there any reason not to copy the codegen api so that we have consistency throughout the code?`

Currently the `__init__` for ODEFunctionGenerator is
```
    def __init__(self, right_hand_side, coordinates, speeds, constants,
                 mass_matrix=None, coordinate_derivatives=None,
                 specifieds=None, linear_sys_solver='numpy',
                 constants_arg_type=None, specifieds_arg_type=None):
```
Coordinates and speeds could be moved to *args like you suggested. I think the key inputs here would then be `right_hand_side, *args, mass_matrix=None, coordinate_derivatives=None` not including the other arguments we've discussed elsewhere in the PR (like `output_eqns`)

jbm950:
```
eom1 = eombase.EOM(coordinates, RHS, speeds=speeds, output_eqns=out_eqns)
eom2 = eombase.EOM(states, forcing_full, mass_matrix=mm_full, num_coordinates=2)
eom3 = eombase.EOM(states, f, mass_matrix=mm, kinematics=G)
```
This is what I was imagining where the speeds karg lets the class know explicitly the form the user input.

jbm950: In keeping with the codegen api though the argument `kinematics` would need to be changed to `coordinate_derivatives`

examples/simple_pendulum/run.py
jbm950: I was imagining it being the kinetic energy of each individual body since you said that a benefit of the `*Methods` is that they contain extra info such as a list of each of the bodies which in turn keeps track of its kinetic energy.

moorepants: Ok, that could work. You need to then have a way for the user to know which body/particle that the rows/columns of the array correspond to. And the kinetic energy of the whole system is important too. Would the user just sum along the axes of the numpy array for the total kinetic energy or would you precalculate that an it would be an extra row/col? This is where it would be good to show the desired output.

examples/two_mass_spring_damper/run.py
jbm950: I agree that lagrange's method uses q, q' and q'' and doesn't even store a "speeds" variable (`LagranesMethod.u` returns `LagrnagesMethod._qdots`). This results in `Derivative(q(t), t)` potentially showing up in `L.mass_matrix` and `L.forcing`. I suppose these would need to be replaced with `u`'s so that no derivative terms show up in either `L.mass_matrix` or `L.forcing`. Leaving the equations in terms of q'' does not have any other functional difference that I can think of because `LagrangesMethod.mass_matrix_full` and `LagrangesMethod.forcing_full` specify that u=q' and u' or q'' doesn't actually show up in the resulting symbolics.

moorepants: Good to know. We need to figure out how to handle this. Lagrange's method should have a way to covert from the second order form to first order form. Kane's method does that explicitly from the beginning by specifying the kinematic differential equations.

examples/two_mass_spring_damper/run.py
jbm950: `Since everything is a variable in the equations of motion, this should map a symbol to an expression instead of mapping a string to an expression. That seems to be more consistent.`

Are you meaning something along the lines of

```python
KE = symbols('KE')
out_eqns = {KE: expr}
```

`Also, should we allow simply an iterable of expressions too? That's the way most things are passed into the *Methods classes, for example.`

Are you meaning

```python
[1]: out_eqns = {'Kinetic Energy': [expr1, expr2, ...]}
```

or 

```python
[2]: out_eqns = [expr1, expr2, ...]
```

The reason I chose to use a string is so each equation could be differentiated by a distinct label. This would be useful when there are a variety of types of output equations being used as each one would be able to be labeled distinctly its name of the equation (and could later be pulled by name).

moorepants: Most things are passed into the `*Method`s classes as:

```python
exprs = [expr1, expr2, expr3, ...]
```

Where `expri` represents the left hand side of:

```
expri = 0
```

In the `System` class we use mostly dictionaries to map sympy symbolics to numerics (floats or arrays of floats). `System` was designed years after `*Method`s and has some nice advantages. So in `System` you could do something like:

```python
sys.plot_state(q1, q2, q3, u1)
```

Where q1, q2, q3, u1 are SymPy symbols. They can be easily printed in the plot using LaTeX. But you may also sometimes what to know what a state is, i.e. q1 is the angle angle between the arm of the pendulum and the ground...

If you use symbols for the output equations you can do:

```python
sys.plot(q1, q2, ke)
```

Where you can know include other symbols that represent the output equations.

I don't know what the best idea for all these API choices are, keep in mind I'm just throwing out ideas and trying to insert some of the historical information about the API design.

examples/two_mass_spring_damper/run.py
jbm950: `Why is num_coordinates needed? And how do you know what order the coordinates are in wrt to the state vector?` 

Number of coordinates is needed for the `sys.plot_coordinates()` method that was discussed last Friday as being a potentially useful method for system. It would be assumed that state is held in the following format [coordinates, speeds, other (ex. Lagrange coefficients)] in which case knowing the number of coordinates the code can pick that many elements off the beginning of the states iterable and have just the coordinates.

`Also, can this work`

This would not be the ideal method because then EOM would have to account for and prepare for every single method that could be used. This would be a maintenance nightmare and make adding new methods an annoying difficult process which is the opposite of the desired result.

`Or should this be how you do it:`

This would be one method. In this way each `*Method` can organize its results into the format that is expected by EOM and would make adding new methods easier as they would simply form the equations of motion their own way then format them for the `EOM` class.

The other way would be `class *Method(eom_base):`. In this method the equations of motion will still need to be specified to the universal API where the base class would provide a blue print. This would require more faith in the programmer of the new `*Method` class to understand and mimic the desired API where the returned object would automatically have the desired API. 

Ex.
```python
class eom_base(object):
    @property
    def mass_matrix(self):
        return self._m_d
```
would require the designers of new classes to include `self._m_d` in their definitions. The advantages of subclassing is that the different components of `*Methods` remain with the equation of motion data (ex. `auxiliary_eqns` or `solve_multipliers()`). For our purposes though I think a returned `EOM` instance might be better as the user can still use the additional information as the `*Method` instance hasn't been deleted, it just isnt what is passed to system.

`I think we need a way to pass in the bodies and loads too.`

I agree this can be useful information.

examples/two_mass_spring_damper/run.py
jbm950: `Most things are passed into the *Methods classes as:... expri = 0`

I think that the form `out_eqns = {'Kinetic Energy': [expr1, expr2, ...]}` is a good combination of the previous API where `expri = 0` still holds and you can differentiate between the basic equations you put in. I think `out_eqns= [expr1, expr2, expr3, ...]` would be too vague because there is a variety of different output equations you could want to specify (for example expr1 = kinetic energy of body 2, expr2 = contact forces in joint3, expr4 = auxiliary equation).

`In the System class we use mostly dictionaries ... Where you can now include other symbols that represent the output equations.`

Might it be that strings or symbols could be used for the keys in the dictionary? I think strings might be simpler because some of the output equations could be specified by the `*Methods` class in the `*Methods.to_eom()` method (like Kane's auxiliary equations) and the user would have to come up with the symbol that the method used instead of just a string with the label for the expression (which brings up the need to be able to pass additional equations to `output_eqns` outside of `EOM.__init__(...)`).

examples/two_mass_spring_damper/run.py
jbm950: `What does this print?`

I was thinking something like `{'kinetic energy': np.array([...])}`

`This shows the desired input and output,`

This is a good idea. I'll set up a page with the three examples that have been used so far.

moorepants: You might as well do this in the sphinx documentation, because once everything is implemented these will simply become doctests.

jbm950:
I have started a documentation page where I have placed all three examples. I am currently changing them from their code form to documentation format and have then I will begin addressing the changes we discussed above.

docs/eom_base.rst
```diff
@@ -0,0 +1,233 @@
...
+
+    >>> plt.plot(sys.times, out[:, 1])  
+    >>> plt.plot(sys.times, out[:, 2])
+    >>> plt.show()
+
+System also has the ability to display the trajectories of the coordinates or
+states without you having to call matplotlib yourself. ::
+
+    >>> sys.plot_coordinates()
+    >>> sys.plot_states()
```
moorepants: How about a `plot_trajectories` method that works like so:

```python
>>> sys.plot_trajectories(*sys.coordinates)
>>> sys.plot_trajectories(sys.coordinates, sys.speeds)  # makes sub plots one for coords one for speeds
>>> sys.plot_trajectories((q1, q2), (u1, u2))  # also two subplots
>>> sys.plot_trajectories(q1, q2, ke)  # takes any symbol that system knows about, ke would have to be the key hen defining the output equations
>>> sys.plot_trajectories(c * u + k * q)  # we could even let the user plot an expression of symbols that are known to system
```

jbm950:
Your plot trajectories comment has been addressed along with the other items we've discussed thus far.

docs/eom_base.rst
```diff
@@ -0,0 +1,255 @@
...
+    >>> out
+    array([[1.0, 0.0, 0.0, 0.0], [...], ...])
+
+The user specified output equations should be able to be determine now that the
+time simulation of the states is complete. ::
+
+    >>> sys.output_eqns()
+    {"kinetic energy": array([[0.0], [...], ...])}
+    >>> eom1.output_eqns_results
+    {"kinetic energy": array([[0.0], [...], ...])}
```
moorepants: I think that you may be able to allow any key here, str, Symbol, etc. And then when the user asks for a specific symbol it will search the coordinates, speeds, specified, and parameters (all sympy symbols) and then finally it will look for the key in the output_equations dictionary, which can be anything. If it is a sympy symbol, the plotting will render it with LaTeX, but if it is any other arbitrary key it will print as whatever the `__str__` method of that object produces. So this if fine, but I think the default should be that you pass in a SymPy symbol, because currently System is only aware of SymPy symbols and functions. In the far future we could imagine having a specialized symbol/function that did this:

```python
>>> q = DynamicSymbol('q', name='Angle between ground and arm', unit=radian)
>>> q
q(t) rad
>>> q.name
Angle between ground and arm.
>>> q.unit
radian
>>> theta = DynamicSymbol('theta', unit=degree)
>>> q + theta
q(t) + theta(t) * pi / 180 rad
>>> ke = DynamicSymbol('K_e', name='Kinetic energy of the arm')
>>> output_equations = {ke: 1 / 2 * I * q.diff() ** 2} 
```

This would then allow us to have a rich way to print all these objects in the various visualizations. But for now we don't have anything but SymPy symbols to work with. The name and units are not yet part of the whole design, so we should keep things consistent.

jbm950: I could add a second output equation to the dictionary to show that strings and symbols are potential keys.

moorepants: Sounds good, These will eventually  be translated to unit tests that will ensure it can work with a variety of inputs.

jbm950:
I have put my notes that I have regarding equations of motion "forms" that I have found so far on my [website](http://jbm950.github.io/notes/dynamics/dynamics_main.html). Most sources that I have found are only bothered with forming the equations of motion and make no mention of different arrangements and what the names of those arrangements might be. I'm not sure that there even are names for the different arrangements and we may even be free to name them ourselves for our software.

jbm950:
I have written the rest of the public attributes and methods into the example. If there are any I missed or didn't think of let me know. I'm going to go through and add common error messages to specify invalid input combinations or attribute accesses when inputs aren't passed.

jbm950:
@angadhn or @moorepants could someone do a review of the .rst file? Also I think the check failed because of a change to numpy.

oliverlee:
I've merged the fix so feel free to merge/rebase this branch.

docs/eom_base.rst
```diff
@@ -0,0 +1,372 @@
...
+
+This example walks through the creation of a two mass spring damper system,
+manually entered to the `eombase` class and numerically simulated using
+`pydy.system.System`. ::
+
+    >>> import matplotlib.pyplot as plt
+    >>> from numpy import linspace
+    >>> from pydy.system import System
+    >>> from sympy import symbols, Matrix
+    >>> from sympy.physics.mechanics import dynamicsymbols, eombase
```
moorepants: `>>> from sympy.physics.mechanics.eombase import EOM`

Symbolic Equations Of Motion: EoM, EOM, EquationsOfMotion

`*Methods` classes will generate equations of motion.

moorepants: `from . import EquationsOfMotion as EOM`

docs/eom_base.rst
```diff
@@ -0,0 +1,372 @@
...
+    >>> PE = symbols("PE")
+    >>> out_eqns = {"kinetic energy": KE, PE: 0.5*k1*x1**2 + 0.5*k2*(x2-x1)**2}
+
+The `eombase` instances are now ready to be initialized. To showcase possible
+input combinations, all three of the equations of motion forms will be used. ::
+
+    >>> eom1 = eombase.EOM(coordinates, RHS, speeds=speeds, 
+    ...                    output_eqns=out_eqns)
+    >>> eom2 = eombase.EOM(states, f_full, mass_matrix=mm_full,
+    ...                    num_coordinates=2)
```
moorepants: I don't think `num_coordinates` is sufficient. For example, what if the states are in this order:

`q1, q2, u3, u2, q3, u1`

We should allow any order of the states, as users that manually generate them might want an arbitrary  order.

The other thing to note is that `q_i` does not necessarily correspond to `u_i` in the general sense and `len(q)!=len(u)`. So, there needs to be a way to tell the class which indices are coordinates or speeds.

Maybe `coord_idxs=(3, 2, 5, 9)`.

docs/eom_base.rst
```diff
@@ -0,0 +1,372 @@
...
+
+    >>> KE = 0.5 * (m1*u1**2 + m2*u2**2)
+    >>> PE = symbols("PE")
+    >>> out_eqns = {"kinetic energy": KE, PE: 0.5*k1*x1**2 + 0.5*k2*(x2-x1)**2}
+
+The `eombase` instances are now ready to be initialized. To showcase possible
+input combinations, all three of the equations of motion forms will be used. ::
+
+    >>> eom1 = eombase.EOM(coordinates, RHS, speeds=speeds, 
+    ...                    output_eqns=out_eqns)
```
moorepants: I like the optional speeds so that the first arg is the states.

docs/eom_base.rst
```diff
@@ -0,0 +1,372 @@
...
+is assuming that the equations of motion have been put in in form [2]. If the
+equations of motion are input in form [1], the class will not have a notion of
+the mass matrix or the forcing vector and as such will return errors if there
+is an attempt to access these attributes. ::
+
+    >>> eom4 = eombase.EOM(states, f, coordinate_derivatives=G)
+    SyntaxError: Need to specify a mass matrix for eom form [3]
+    >>> eom1.mass_matrix
+    AttributeError: Mass matrix is not specified for eom form [1]
+    >>> eom1.mass_matrix_full
```
moorepants: forcing ->  right_hand_side
mass_matrix -> mass_matrix

mass_matrix_full -> mass_matrix_first_order
forcing_full -> right_hand_side_first_order

mass_matrix -> dynamic_mass/inertial_matrix
kinematic_matrix
composite_mass_matrix, composite_matrix

# Dynamic Differential Equations
## Implicit
M u' = F
## Explicit:
u' = inv(M) * F

# Kinematic Differential Equations
## Implicit
H q' = G
## Explicit
q' = inv(H) * G

# Combined Differential Equations
## Implicit
W x' = Y
## Explicit
x' = inv(W) * Y

moorepants: forcing ->  right_hand_side
mass_matrix -> mass_matrix

mass_matrix_full -> mass_matrix_first_order
forcing_full -> right_hand_side_first_order

mass_matrix -> dynamic_mass/inertial_matrix
kinematic_matrix
composite_mass_matrix, composite_matrix

# Dynamic Differential Equations
## Implicit
M u' = F
## Explicit:
u' = F* = inv(M) * F

# Kinematic Differential Equations
## Implicit
H q' = G
## Explicit
q' = G* = inv(H) * G

# Combined Differential Equations
## Implicit
W x' = Y
## Explicit
x' = Y* = inv(W) * Y

F, G, and Y are the right hand side of the implicit forms of the * differential equations.
M, H, and W are the (inertial) coefficient matrices of the implicit forms of the * differential equations.
F*, G*, and Y* are the right hand side of the explicit forms of the * differential equations.

moorepants: F: `.dyn_implicit_rhs`
G: `.kin_implicit_rhs`
Y: `.comb_implicit_rhs`
M: `.dyn_implicit_mat`
H: `.kin_implicit_mat`
W: `.comb_implicit_mat`
F*: `dyn_explicit_rhs`
G*: `kin_explicit_rhs`
Y*: `.comb_explicit_rhs`

docs/eom_base.rst
```diff
@@ -0,0 +1,372 @@
...
+    >>> out
+    array([[1.0, 0.0, 0.0, 0.0], [...], ...])
+
+The user specified output equations should be able to be determined now that the
+time simulation of the states is complete. These equations are calculated by
+sys.output_eqns and so if there is an attempt to see the results first an error
+will be returned. ::
+
+    >>> eom.output_eqns_results
+    ValueError: The output equations first need to be calculated
```
moorepants: Both integrate() and evaluating the output equations require computation (sometime a lot), so they should be methods. but if we want to plot the trajectories, then it would be convenient to have them stored in System. Or the functions on System like `.plot()` should take the trajectories (a numpy array) as an argument. I'm not sure what is best. Or we can return a result object that has the plot methods.

docs/eom_base.rst
moorepants: Would 'SymbolicSystem" be a better name than `EquationsOfMotion`, being that we include bodies and loads in addition to the questions.

docs/eom_base.rst
```diff
@@ -0,0 +1,372 @@
...
+    >>> sys.output_eqns()
+    {"kinetic energy": array([[0.0], [...], ...]),
+     PE: array([[5.0], [...], ...])}
+    >>> eom1.output_eqns
+    {"kinetic energy": KE, PE: 0.5*k1*x1**2 + 0.5*k2*(x2-x1)**2}
+    >>> eom1.output_eqns_results
```
moorepants: How about `output_trajectories` and `state_trajectories` for these attributes. Also, we should provide the state and output trajectories in the same form, either a dictionary mapping keys to 1d arrays, 2d arrays containing a row for each state/output, or both. The 2d array form is quite convenient for plotting purposes and more efficient gross numerical computations with numpy. If we wanted a new dependency we could return a pandas DataFrame which is like a dict but has all kinds of niceties for time series analysis.

docs/eom_base.rst
```diff
@@ -0,0 +1,372 @@
...
+
+    >>> plt.plot(sys.times, out[:, 1])  
+    >>> plt.plot(sys.times, out[:, 2])
+    >>> plt.show()
+
+System also has multiple plotting capabilities built into the class. ::
+
+    >>> sys.plot_coordinates()
+    >>> sys.plot_speeds()
+    >>> sys.plot_states()
```
moorepants: All above should call line 132 internally I suppose.

docs/eom_base.rst
```diff
@@ -0,0 +1,372 @@
...
+
+Lastly here are some attributes accessible from the `EOM` class. ::
+
+    >>> eom1.states
+    (x, y, u, v, lam)
+    >>> eom2.coordinates
+    (x, y)
+    >>> eom3.speeds
+    (u, v)
+    >>> eom1.rhs
```
moorepants: Is `right_hand_side` preferable.

docs/eom_base.rst
```diff
@@ -0,0 +1,372 @@
...
+    >>> eom2.mass_matrix_full
+    Matrix([[1, 0, 0, 0, 0], [0, 1, 0, 0, 0], [0, 0, 1, 0, -x/m], [0, 0, 0, 1,
+             -y/m], [0, 0, 0, 0, l**2/m]])
+    >>> eom3.forcing
+    Matrix([0, 0, u**2 + v**2 - g*y])
+    >>> eom3.mass_matrix
+    Matrix([[1, 0, -x/m], [0, 1, -y/m], [0, 0, l**2/m]])
+    >>> eom1.alg_con
+    [4]
+    >>> eom1.dynamic_symbols()
```
moorepants: Why a method? Should it be an attribute?

moorepants:
@oliverlee We are working on this API here at SciPy this week. Would you have some time to comment soon? We'd like to home in on a nice API that can describe a rigid body system's symbolic equations of motion in a general way.

docs/eom_base.rst
moorepants: I wonder if we could do something like:

EquationsOfMotion.implicit.dynamic.rhs
EquationsOfMotion.implicit.dynamic.mat

or

EquationsOfMotion.explicit_dynamic().rhs

moorepants:
@jbm950 I'm happy with moving forward with this. I've made comments about things I think should be adjusted. There are some details about naming but I think moving forward with the idea about having a symoblic system description object is good. All of the Method objects can then return this object as a primary result.

docs/eom_base.rst
jbm950: For the inputs to the SymbolicSystem initialization I'm leaving mass_matrix and coordinate_derivatives to both match the current api and because mass_matrix can take in either dyn_implicit_mat or comb_implicit_mat. I'll change the attributes to match the new names.

docs/eom_base.rst
jbm950: I do not quite understand the 2d array form you have described. Since this is more related to pydy's System object now I'm going to move forward with SymbolicSystem in SymPy.

moorepants: That's fine.

moorepants:
This is looking good. The only thing I'm not sure is addressed is the issue associated with the fact that users may not order there states in any particular order. There was some assumption that `states = coordinates + speeds`. A user that manually populates the state vector may order them in any fashion, for example: `[u2, q3, u1, q2, q1]`. So maybe there should be two attributes that list the indices of the coordinates and speeds, in this case: `q_idxs = [4, 3, 1]` and `u_idxs = [2, 0]`.

jbm950:
```python
>>> symsystem2 = system.SymbolicSystem(states, comb_implicit_rhs, 
...                                    mass_matrix=comb_implicit_mat,
...                                    alg_con=alg_con_full,
...                                    coord_idxs=(0, 1))
>>> symsystem3 = system.SymbolicSystem(states, dyn_implicit_rhs, 
...                                    mass_matrix=dyn_implicit_mat,
...                                    coordinate_derivatives=kin_explicit_rhs,
...                                    alg_con=alg_con, coord_idxs=(0, 1), 
...                                    speed_idxs=(2, 3))
```

Isn't that covered by this code?

moorepants:
Ok, cool, must have missed it.

PR #356, opened 2016-07-14
kwargs can be passed from System.generate_ode_function to the matrix generator

moorepants:
This implements an _options attribute to the subclass of ODEFunctionGenerator
and thus allows various options needed for a specific MatrixGenerator to be
supplied all the way at the System.generate_ode_function and
generate_ode_function levels.

  commit message.
  directory)
  follow deprecation cycles.)

@asmeurer This fixes the issue you had this morning.

### DIFF
--- a/pydy/codegen/ode_function_generators.py
+++ b/pydy/codegen/ode_function_generators.py
@@ -641,6 +641,13 @@ class CythonODEFunctionGenerator(ODEFunctionGenerator):
 
     def __init__(self, *args, **kwargs):
 
+        self._options = {'tmp_dir': None,
+                         'prefix': 'pydy_codegen',
+                         'cse': True,
+                         'verbose': False}
+        for k, v in self._options.items():
+            self._options[k] = kwargs.pop(k, v)
+
         if Cython is None:
             raise ImportError('Cython must be installed to use this class.')
         else:
@@ -648,9 +655,12 @@ class CythonODEFunctionGenerator(ODEFunctionGenerator):
 
     __init__.__doc__ = ODEFunctionGenerator.__init__.__doc__
 
-    @staticmethod
-    def _cythonize(outputs, inputs):
-        return CythonMatrixGenerator(inputs, outputs).compile()
+    def _cythonize(self, outputs, inputs):
+        g = CythonMatrixGenerator(inputs, outputs,
+                                  prefix=self._options['prefix'],
+                                  cse=self._options['cse'])
+        return g.compile(tmp_dir=self._options['tmp_dir'],
+                         verbose=self._options['verbose'])
 
     def _set_eval_array(self, f):
 
--- a/pydy/system.py
+++ b/pydy/system.py
@@ -459,8 +459,8 @@ class System(object):
 
     def generate_ode_function(self, **kwargs):
         """Calls ``pydy.codegen.ode_function_generators.generate_ode_function``
-        with the appropriate arguments, and sets the
-        ``evaluate_ode_function`` attribute to the resulting function.
+        with the appropriate arguments, and sets the ``evaluate_ode_function``
+        attribute to the resulting function.
 
         Parameters
         ----------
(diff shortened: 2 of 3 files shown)


pydy/codegen/ode_function_generators.py
```diff
@@ -641,16 +641,26 @@ class CythonODEFunctionGenerator(ODEFunctionGenerator):
 
     def __init__(self, *args, **kwargs):
 
+        self._options = {'tmp_dir': None,
+                         'prefix': 'pydy_codegen',
+                         'cse': True,
+                         'verbose': False}
```
moorepants: This `_options` dictionary could probably be built by inspecting the methods of the `CythonMatrixGenerator`, but I didn't try to figure that out.

moorepants:
@jbm950, would you mind reviewing this?

pydy/codegen/ode_function_generators.py
jbm950: I'll be able to give it more of a look over later today or possibly tomorrow. First look though it appears that you are adding more things that a user can set. If this is the case should this be added to a docstring or documentation somewhere? If I remember right the docstring is built dynamically for this class and may be difficult to edit.

pydy/codegen/ode_function_generators.py
```diff
@@ -641,16 +641,26 @@ class CythonODEFunctionGenerator(ODEFunctionGenerator):
...
+        for k, v in self._options.items():
+            self._options[k] = kwargs.pop(k, v)
+
```
jbm950: This works as a way to pull the arguments out of kwargs properly but ends up setting each item in the `self._options` dictionary twice. Another way would be to just call kwargs for each item like this

```python
self._options = {}
self._options['tmp_dir'] = kwargs.pop('tmp_dir', None)
self._options['prefix'] = kwargs.pop('prefix', 'pydy_codegen')
self._options['cse'] = kwargs.pop('cse', True)
self._options['False'] = kwargs.pop('verbose', False)
```

This method probably a bit more verbose and thus readability is reduced but I imagine the performance would improve because each item in the dictionary is only being set once. What do you think?

jbm950:
Looks like you added input support for four new arguments (`prefix`, `tmp_dir`, `cse` and `verbose`) but only added tests for `prefix` and `tmp_dir`. I understand a test for `verbose` would probably be difficult because it's have to read the output of the function call somehow but would it be difficult to add a test for the `cse` option?

That, the documentation comments and the question on possibly increasing the efficiency are the only things I have. Other than that it looks like the functionality was extended well.

moorepants:
> Looks like you added input support for four new arguments (prefix, tmp_dir, cse and verbose) but only added tests for prefix and tmp_dir. I understand a test for verbose would probably be difficult because it's have to read the output of the function call somehow but would it be difficult to add a test for the cse option?

All of the options are tested in the CythonMatrixGeneratorTests. All that is needed to test here is whether the arguments are passed in from `System.__init__()` to `CythonMatrixGenerator`. The tests I have prove that the args make it there. I don't think it's necessary to check more than one of the optional arguments.

> That, the documentation comments and the question on possibly increasing the efficiency are the only things I have. Other than that it looks like the functionality was extended well.

The docstring of system says that any extra kwargs will be passed through to the underlying routines and those routines have clear docstrings about their optional arguments. I don't think added all of the options for the underlying generators is necessary in System's docstrings. But I'll look into adding something if there is a good spot for it.

jbm950:
That works for me then. What did you think about building the options dictionary one item at a time and getting rid of the call to `enumerate()`?

pydy/codegen/ode_function_generators.py
moorepants: Actually, my code might fail if the options aren't in kwargs. I'll look at this again.

moorepants: Oh, no it is fine. Yeah, I could write it your way too. I'm not sure it really matters, as this is not a place that performance matters.

jbm950:
> I'm not sure it really matters, as this is not a place that performance matters.

In that case readability > performance and I think the way you have it currently written makes it easier to see at a glance what is supposed to be in the `self._options` dictionary.

I'll go ahead and merge this unless you have any objections.

moorepants:
FYI, for next time. Please review the list at the top of the PR and check off things that have been taken care of before merging.

PR #357, opened 2016-07-16
updated circle funtion

remcmurry:
changed function to relate thickness of drawn cylinder to radius object such that thickness will always be 1/50 of circle diameter, 1/25 of radius. such that the thickness and fact that it is a cylinder is not readily apparent to the casual user.

### DIFF
--- a/pydy/viz/static/js/dyviz/scene.js
+++ b/pydy/viz/static/js/dyviz/scene.js
@@ -246,7 +246,7 @@ DynamicsVisualizer.Scene = Object.extend(DynamicsVisualizer, {
             case "Circle":
                 var geometry = new THREE.CylinderGeometry(object.radius,
                                                           object.radius,
-                                                          0.5,100);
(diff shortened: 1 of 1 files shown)


pydy/viz/static/js/dyviz/scene.js
```diff
@@ -246,7 +246,7 @@ DynamicsVisualizer.Scene = Object.extend(DynamicsVisualizer, {
             case "Circle":
                 var geometry = new THREE.CylinderGeometry(object.radius,
                                                           object.radius,
-                                                          0.5,100);
+                                                          object.radius/25,100);
```
remcmurry: this is the line you could enter zero or a higher divisor if you want a thinner disc shape.

remcmurry:
here is a pic of the result, this can be changed to be thinner if we want 
image

if we choose to always use zero the two sides have different colors blue and black in my case.
image

image

PR #368, opened 2017-04-08
Adds implementation for Tube Geometry.

### LINKED ISSUE #367: Visualization of lines between points
Hi,

I would like to know if it is possible to draw lines between points of interest (e.g. visualize muscles). There seems to be a Shape object Tube. Please correct me if I am wrong, but from the js part it is probably unimplemented:

[code](https://github.com/pydy/pydy/blob/master/pydy/viz/static/js/dyviz/scene.js#L281)

Regards,
Dimitris

tarzzz:
Fixes #367 .

This PR adds the implementation for TubeGeometry.

@moorepants Please review. Kindly note that this does not add the `Tube` shape
in `Edit Objects` dropdown in Frontend/UI. That requires some more work, so I will handle it as a separate PR.

### DIFF
--- a/pydy/viz/shapes.py
+++ b/pydy/viz/shapes.py
@@ -786,7 +786,7 @@ class Tube(Shape):
 
     @property
     def points(self):
-        return self._points
+        return self._points.tolist()
 
     @points.setter
     def points(self, new_points):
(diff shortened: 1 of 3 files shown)


moorepants:
Closed in favor of #376.

PR #370, opened 2017-05-28
Removed the explicit march flag for Theano compilation.

moorepants:


moorepants:
This doesn't have the desired effect. Compilation fails on Travis without the march flag.

PR #373, opened 2017-05-28
Update olivers

moorepants:


moorepants:
The old style args parsing has been removed for 0.4.0 deprecation.

PR #374, opened 2017-05-29
Updates code to reflect deprecation of kanes_equations.

moorepants:


pydy/utils.py
```diff
@@ -33,6 +33,12 @@ def sympy_equal_to_or_newer_than(version, installed_version=None):
     return parse_version(v) >= parse_version(version)
 
 
+def sympy_newer_than(version):
+    """Returns true if the installed version of SymPy is newer than the
+    provided version string."""
+    return parse_version(SYMPY_VERSION) > parse_version(version)
```
moorepants: This can probably be removed now because SymPy no longer has the trailing "-git" on the dev version numbers.

moorepants: Since setuptools if far from version 8 and SymPy has had the correct versions for over a year (maybe two). I think all this could be dropped.

PR #459, opened 2022-02-02
Prep for 0.6.0

moorepants:
This PR will get a 0.6.0 release ready.

- [x] Bump all dependencies up to Ubuntu 20.04 versions and drop older things.
- [x] Drop Python 2.7
- [x] Make sure new features are documented


moorepants:
For some reason nose is not picking up all tests on windows py3.8 and py3.9 but it does on py3.7. Switching to pytest is what needs to happen, but I'll wait on that.

PR #462, opened 2022-02-06
Carvallo-Whipple model with lateral tire constraint forces exposed

moorepants:
This example formulates the Carvallo-Whipple non-linear bicycle model and uses auxiliary speeds to expose the ground-wheel reaction forces. It generates Octave code that numerically evaluates the ground reaction forces given kinematic inputs.

cc/ @oliverlee @JeswinKoshyCherian,

This model generates functions for the independent and dependent speed equations and solves them numerically (in Octave). believe this is the correct approach (as opposed to what `pydy.system.System()` currently does in the Carvallo-Whipple model in the docs). If I'm right, then it may remove some of the numerical issues that the example in the docs faces. It would be nice to add CythonMatrixGenerator in addition to the Octave one so that we can work fully in Python and see if the model can correctly simulate without the numerical issues. If it does, then we can build on it for the cargo bike model.


examples/bicycle_wheel_contact_constraint_forces/lateral_tire_forces.m
```diff
@@ -0,0 +1,69 @@
...
+res = Ap_nh\bp_nh;
+u3p = res(1);  % yaw angular acceleration
+u5p = res(2);  % pitch angular acceleration
+u8p = res(3);  % front wheel angular acceleration
+
+% finally calculate the constraint forces
+[A, b] = eval_lat_forces([q4, q5, q7], ...
+                         [u3, u4, u5, u6, u7, u8], ...
+                         [u3p, u4p, u5p u6p, u7p, u8p], ...
+                         const);
```
moorepants, line 64: If we create an "evaluate udots" function before this one, then that can be used in the integrator.

examples/bicycle_wheel_contact_constraint_forces/lateral_tire_forces.m
```diff
@@ -0,0 +1,69 @@
...
+q4 = q(1);  % roll angle
+q7 = q(2);  % steer angle
+
+u4 = u(1);  % roll rate
+u6 = u(2);  % rear wheel rate
+u7 = u(3);  % steer rate
+
+u4p = up(1);  % roll angular acceleration
+u6p = up(2);  % rear wheel angular acceleration
+u7p = up(3);  % steer angular acceleration
```
moorepants, line 31: For a simulation function we need to calculate these with the equations of motion. Something like:

```python
M = kane.mass_matrix
f = kane.forcing
eval_M, eval_f = MatrixGenerator(M, f)
u_independent_dot = np.linalg.solve(M, f)
u4p = u_independent_dot[0]
u6p = u_independent_dot[6]
u7p = u_independent_dot[7]
```

examples/bicycle_wheel_contact_constraint_forces/whipple.py
```diff
@@ -0,0 +1,588 @@
...
+# front wheel contact velocity
+fn.v2pt_theory(fo, N, F)
+
+# create a front "yaw" frame that is equivalent to the A frame for the rear
+# wheel.
+# G['1'] lies in the ground plane and points in the direction of the wheel
+# contact path E['2'] X A['3'] gives this unit vector.
+# G['2'] lies in the ground plane and points perpendicular to the wheel
+# contact path. A['3'] X G['1'] gives this unit vector.
+g1_hat = E['2'].cross(A['3'])
```
moorepants, line 330:
```suggestion
g1_hat = E['2'].cross(A['3']).normalize()
```

moorepants: E2 and A3 are not orthogonal in general.

PR #484, opened 2022-12-05
Adding support for Box shapes to dyviz js

ebranlard:
The python code `pydy.viz.shapes.py` provides support for `Box` shapes but the type was not supported by the javascript code. This pull request add support for this, including the possibility to edit the parameters in the GUI when selecting `Edit Objects`. 

I can try to add documentation and unittests but I will wait for your opinion. 

Below is a sample code to test the feature (it might need a "Shift-F5" to reload the javascript from the browser's cache):
```python
import sympy as sp
from sympy.physics.mechanics import Point, ReferenceFrame
from pydy.viz.scene import Scene
from pydy.viz.shapes import Box
from pydy.viz.visualization_frame import VisualizationFrame

frame = ReferenceFrame('e')
point = Point('A')

box = Box(width=4, height=2, depth=0.5, color='blue', name='testbox')
vizf = VisualizationFrame('vizbox',frame, point, box)
scene = Scene(frame, point, vizf)
scene.constants = {}
scene.states_symbols = [sp.Symbol('x')]
scene.states_trajectories = np.zeros((2,1))
scene.times = np.array([0.,1.0])
scene.display()
```

New Feature
-----------

     commit message.
     directory)
     follow deprecation cycles.)

### DIFF
--- a/pydy/viz/static/js/dyviz/materials.js
+++ b/pydy/viz/static/js/dyviz/materials.js
@@ -1,4 +1,4 @@
-DynamicsVisualizer.Geometries = ['Cube', 'Cylinder',
+DynamicsVisualizer.Geometries = ['Box', 'Cube', 'Cylinder',
                                  'Cone', 'Sphere', 'Circle',
                                  'Plane', 'Tetrahedron',
                                  'Octahedron', 'Icosahedron',
--- a/pydy/viz/static/js/dyviz/param_editor.js
+++ b/pydy/viz/static/js/dyviz/param_editor.js
@@ -119,6 +119,12 @@ DynamicsVisualizer.ParamEditor = Object.extend(DynamicsVisualizer, {
                 updated_object.length = jQuery("#_length").val()
                 break;
 
+            case "Box":
+                updated_object.width = jQuery("#_width").val()
+                updated_object.height = jQuery("#_height").val()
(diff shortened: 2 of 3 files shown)


moorepants:
Is there a reason "Cube" isn't sufficient?

moorepants:
There does seem to be a BoxGeometry in Three.js: https://threejs.org/docs/#api/en/geometries/BoxGeometry 

I'm not seeing a CubeGeometry in the three.js docs.

ebranlard:
Having the support for boxes is convenient to visualize geometries that have different lengths in all three directions (like a ladder, a cart,  etc.). To some extent, supporting Cube is redundant, but offers some convenience to the user. 

The python code already have support for both Cube and Box (see  [here](https://github.com/pydy/pydy/blob/master/pydy/viz/shapes.py#L260)), and Cube is actually just implemented as a Box. 

The problem is that the current Javascript code of `pydy` that does not support `Box` but Three.js does support it. 

You are right that in my code change, I've used `CubeGeometry`, and not `BoxGeometry`, which is a bit confusing. The former is an alias for the later (see [here](https://stackoverflow.com/a/51591365/627911)) so the code works as intended.

Since `CubeGeometry` is getting depreciated, we can use `BoxGeometry` both for `Cube` and `Box` in the javascript code. I can update the code if you agree.

PR #508, opened 2025-03-28
Add example of bicycle tire slip model.

moorepants:


Peter230655:
I ran it on my PC, also got the animation going.
But frankly, I do not know what I am looking at.
_Could you give me some hints?_
**Thanks!**

PR #530, opened 2025-09-22
Add optional symjit generator.

moorepants:
This adds a new ODEFunctionGenerator that uses symjit as sympy->numeric code generator and compiler. symjit does not support iterable of iterables as function arguments and it can only output flat arrays, so a fair amount of input/output conversion is needed that slows down both the code generation and the numerical evaluation. Replacing symbols in the equations of motion does not seem to slow symjit's code generation which is clearly very fast but the numeric manipulation and type conversion does slow down the evaluation. If symjit can manage some things faster internally to manage similar inputs that lambdify accepts, then symjit may have equal evaluation speed to the Cython generation code and generate extremely faster.

Here is the benchmark run with 16 pendulum links, 100 integration steps, and 2 second duration.

```
bin/benchmark_pydy_code_gen.py 16 100 2
```

benchmark-results

This needs a bit more tidying and management of the optional dependency before merging.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -34,7 +34,7 @@ jobs:
           channels: conda-forge
       - name: Install basic dependencies
         run: |
-          conda install -q -y setuptools coverage cython matplotlib packaging pytest pytest-cov numpy pythreejs sympy scipy
+          conda install -q -y setuptools coverage cython matplotlib packaging pytest pytest-cov numpy pythreejs symjit sympy scipy
       - name: Install window dependencies
         if: ${{ matrix.os=='windows-latest' }}
         run: |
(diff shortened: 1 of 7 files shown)


Peter230655:
At the very bottom of this: https://github.com/siravan/symjit/issues/3
there is described a way to make symjit considerably faster.
image
I have not seen this im the documentation, yet.
I use it in my comparisons.

moorepants:
Doesn't look like symjit 2.5 is available for Python 3.9 from conda forge.

Peter230655:
> > Just so I am very clerar: Did you use numerical or symbolic inversion here?
> 
> I don't think I will answer, so that you read the code here :)

I read it for 30 min, but still not clear. :-((

PR #548, opened 2025-10-14
explorer1 anomaly

Peter230655:
This simulates the fact, first found out with Explorer1, that the rotation of non rigid bodies may be unstable around their axis of minimal inertia.
- use ``time_first`` keyword
- integrate ``rhs_gen`` into solve_ivp, no gradient function used
- use auxiliary frames in ``PyDy Visualization`` to orient the cylinders use in the Y - direction

Note: 
I have a very similar simulation in pst-notebooks, but if you agree, I'd prefer it to be here and removed from pst-notebooks:
- I think, the more important simulations should be here. This simulation shows, that today anybody can create and solve the eoms, where a few decades ago learned papers were written about.
- in pst-notebooks the PyDy Visualization does not work, and the matplotlib substitute is much inferior


moorepants:
When I check the error message it tells me exactly what the problem is:

```

Extension error:
Cell raised uncaught exception:
TypeError                                 Traceback (most recent call last)
File ~/work/pydy/pydy/pydy/codegen/ode_function_generators.py:1061, in generate_ode_function(*args, **kwargs)
   1059 try:
   1060     # See if user passed in a custom class.
-> 1061     g = generator(*args, **kwargs)
   1062 except TypeError:
   1063     # See if user passed in a string.

TypeError: 'str' object is not callable

During handling of the above exception, another exception occurred:

ImportError                               Traceback (most recent call last)
Cell In[26], line 1
----> 1 rhs_gen = generate_ode_function(
      2     force,
      3     kane.q,
      4     kane.u,
      5     constants=constants,
      6     mass_matrix=mass_matrix,
      7     specifieds=specified,
      8     coordinate_derivatives=kin_eqs_solved,
      9     generator='symjit',
     10     linear_sys_solver='numpy',
     11     constants_arg_type='array',
     12     specifieds_arg_type='array',
     13     time_first=True,
     14 )

File ~/work/pydy/pydy/pydy/codegen/ode_function_generators.py:1066, in generate_ode_function(*args, **kwargs)
   1064 try:
   1065     Generator = generators[generator]
-> 1066     g = Generator(*args, **kwargs)
   1067 except KeyError:
   1068     msg = '{} is not a valid generator.'.format(generator)

File ~/work/pydy/pydy/pydy/codegen/ode_function_generators.py:934, in SymjitODEFunctionGenerator.__init__(self, *args, **kwargs)
    932 symjit_version = metadata.version('symjit')
    933 if parse_version(symjit_version) < parse_version('2.5.0'):
--> 934     raise ImportError('Symjit >= 2.5.0 is required.')
    936 self._options = {'cse': True}
    938 for k, v in self._options.items():

ImportError: Symjit >= 2.5.0 is required.
make: *** [Makefile:20: html] Error 2
Error: Process completed with exit code 2.
```

Peter230655:
I understood it had something to do with symjit not being present. 
But my simulation worked with other checks, and it also worked locally.
So, I thought the error may be related to the check, rather than to my simulation (?)

Peter230655:
I updated my branch. Still seems the same error is there.
Should I switch my simulation to cython to avoid this errror?

moorepants:
Switching to cython or lambdify is the quickest fix. Otherwise symjit needs to be added to the docs build configuration.

moorepants:
This example is not included in the table of contents or the example table.

docs/examples/explorer1_anomaly.rst
```diff
@@ -0,0 +1,756 @@
...
+  as the constants :math:`k_\textrm{{torque}}` and
+  :math:`\mu_\textrm{{torque}}` are courtesy *Dr. David Levinson* via
+  *Dr. Carlos Roithmayr*.
+- The equations of motion are derived using Kane's method.
+- As there is dampening, the total energy of the system decreases.
+- As there are no external forces or torques, the angular momentum of the
+  system must be constant.
+- For the determination of the reaction forces at the connection points the
+  accelerations are needed. As generate_ode_function needs C - contiguous
+  arrays if generator='cython' is used, a bit of care must be taken.
```
moorepants: Either the function isn't being used correctly or there is a bug we need to fix. We shouldn't have to explain anything about array striding in PyDy examples.

docs/examples/explorer1_anomaly.rst
```diff
@@ -0,0 +1,756 @@
...
+- With ``solve_ivp`` some methods, e.g. Radau, return non C-contiguous arrays.
+  Then rhs_gen cannot be used in solve_ivp directy, but a function must be
+  defined which converts the array accordingly.
```
moorepants: Let's avoid this. It is just confusing to a new user to explain such things. This seems to only be the result of using the Radua integrator. Best to use the default integrator, as that should be sufficient.

docs/examples/explorer1_anomaly.rst
```diff
@@ -0,0 +1,756 @@
...
+    import numpy as np
+    import sympy as sm
+    import matplotlib.pyplot as plt
+    import sympy.physics.mechanics as me
+    import time
+    from scipy.integrate import solve_ivp
+    from scipy.optimize import root
+    from scipy.interpolate import CubicSpline
+    from pydy.codegen.ode_function_generators import generate_ode_function
+    from mpl_toolkits.mplot3d import Axes3D  # noqa: F401
```
moorepants: Remove if not used.

docs/examples/explorer1_anomaly.rst
```diff
@@ -0,0 +1,756 @@
...
+
+    import numpy as np
+    import sympy as sm
+    import matplotlib.pyplot as plt
+    import sympy.physics.mechanics as me
+    import time
+    from scipy.integrate import solve_ivp
+    from scipy.optimize import root
+    from scipy.interpolate import CubicSpline
+    from pydy.codegen.ode_function_generators import generate_ode_function
```
moorepants, line 114: Ideally we make use of System() in these examples.

docs/examples/explorer1_anomaly.rst
```diff
@@ -0,0 +1,756 @@
...
+================
+Explorer Anomaly
+================
+
+.. note::
+
+   You can download this example as a Python script:
+   :jupyter-download:script:`explorer1_anomaly` or Jupyter notebook:
+   :jupyter-download:notebook:`explorer1_anomaly`.
+
```
moorepants, line 11: This example really needs a figure detailing all the coordinates, forces, etc.

docs/examples/explorer1_anomaly.rst
```diff
@@ -0,0 +1,756 @@
...
+The solution must be sorted so that it corresponds to KM.q
+
+.. jupyter-execute::
+
+    loesung = sm.solve(kd, [q_ind[i].diff(t) for i in range(len(q_ind))])
+    schluessel = [i.diff(t) for i in kane.q]
+    kin_eqs_solved = sm.Matrix([loesung[i] for i in schluessel])
+
+``generate_ode_function`` does not want the expanded form of the mass matrix
```
moorepants: What is the purpose of this statement? To teach something? Or is is noted a bug or missing feature?

docs/examples/explorer1_anomaly.rst
```diff
@@ -0,0 +1,756 @@
...
+    specified = None
+    constants = np.array(pL)
+
+The solution must be sorted so that it corresponds to KM.q
+
+.. jupyter-execute::
+
+    loesung = sm.solve(kd, [q_ind[i].diff(t) for i in range(len(q_ind))])
+    schluessel = [i.diff(t) for i in kane.q]
+    kin_eqs_solved = sm.Matrix([loesung[i] for i in schluessel])
```
moorepants: The kinematical differential equations can be accessed from the KanesMethod object and they will be in the expected order. This shouldn't be necessary.

docs/examples/explorer1_anomaly.rst
```diff
@@ -0,0 +1,756 @@
...
+
+``generate_ode_function`` does not want the expanded form of the mass matrix
+
+.. jupyter-execute::
+
+    mass_matrix = me.msubs(kane.mass_matrix, {i: 0 for i in aux})
+    force = me.msubs(kane.forcing, {i: 0 for i in aux + F_r})
```
moorepants: This shouldn't be necessary and may point to a bug in SymPy.

docs/examples/explorer1_anomaly.rst
```diff
@@ -0,0 +1,756 @@
...
+        atol=1.e-8,
+        rtol=1.e-8,
+        )
+
+    resultat = resultat1.y.T
+    print('resultat shape', resultat.shape)
+    print(resultat1.message, '\n')
+
+    print(f"To numerically integrate an interval of {interval} sec the "
+          f"routine cycled {resultat1.nfev:,} times")
```
moorepants: If you were to use the System() class you can avoid much of the book keeping and extra code in these simulation sections.

docs/examples/explorer1_anomaly.rst
```diff
@@ -0,0 +1,756 @@
...
+    ax[0].set_ylabel('Energy [J]')
+    ax[0].set_title('Energy of the system')
+    _ = ax[0].legend()
+
+    max_x = np.max(ang_momentum_lam(*(resultat.T), *pL_vals)[0])
+    max_y = np.max(ang_momentum_lam(*(resultat.T), *pL_vals)[1])
+    max_z = np.max(ang_momentum_lam(*(resultat.T), *pL_vals)[2])
+    min_x = np.min(ang_momentum_lam(*(resultat.T), *pL_vals)[0])
+    min_y = np.min(ang_momentum_lam(*(resultat.T), *pL_vals)[1])
+    min_z = np.min(ang_momentum_lam(*(resultat.T), *pL_vals)[2])
```
moorepants: Can `ang_momentum_lam` be called a single time and the values extracted from that single result?

docs/examples/explorer1_anomaly.rst
```diff
@@ -0,0 +1,756 @@
...
+        ax[1].set_title('Angular momentum of the system')
+    _ = ax[1].legend()
+
+
+Calculate Reaction Forces on Points, where the Antennas are attached to
+Explorer
+
+Calculate the accelerations needed for the reaction forces. rhs_gen needs
+C - contiguous arrays if ``generator='cython'`` is used,
+so the inputs have to be converted here accordingly.
```
moorepants: This should not be necessary and points to a bug somewhere. (or misuse)

moorepants:
I'm open to adding this example. It is long but reasonably tame in terms of complexity.

Peter230655:
> This example is not included in the table of contents or the example table.

I am unsure I understand what you mean.

docs/examples/explorer1_anomaly.rst
Peter230655: > Can `ang_momentum_lam` be called a single time and the values extracted from that single result?

Dumb programming on my part!
Corrected.

docs/examples/explorer1_anomaly.rst
Peter230655: > The kinematical differential equations can be accessed from the KanesMethod object and they will be in the expected order. This shouldn't be necessary.

I did not realize this.
But if they are available, why would ``generate_ode_function`` ask for them?

moorepants:
> I am unsure I understand what you mean.

There is no way for a person to navigate to your page like all the other examples.

docs/examples/explorer1_anomaly.rst
moorepants: It needs every symbolic thing necessary to form the equations of motion.

Peter230655: > It needs every symbolic thing necessary to form the equations of motion.

I agree, but if I can get them from the KanesMethod object why can ``generate_ode_function not get them directly?

moorepants:
```python
s = System(kane)
s.integrate()
```

gets everything from kane automatically (internally it populates generate_ode_function()).

generate_ode_function requires exactly what it asks for and you have to provide it.

docs/examples/explorer1_anomaly.rst
Peter230655: > Either the function isn't being used correctly or there is a bug we need to fix. We shouldn't have to explain anything about array striding in PyDy examples.

I raised issue 23657 with scipy a while ago. They confirmed that ``Radau`` does not return C_contiguous arrays, but as they did not claim so, not their problem. (``BDF`` also does not, others I did not try)
My thinking was: 
If a user has a stiff problem and wants to use ``Radau`` he will get an error message. At leat it took me quite a while to figure out what the problem was, So I thought, such hints would be helpful, at least until ``generate_ode_function`` will handle this internally.

docs/examples/explorer1_anomaly.rst
Peter230655: > Let's avoid this. It is just confusing to a new user to explain such things. This seems to only be the result of using the Radua integrator. Best to use the default integrator, as that should be sufficient.

I do not recall, whether the default integrator worked here. It is a problem at least with Radau and with BDF.
We should probably correct it within ``generate_ode_function```. Based of the reply on my issue 23657 with scipy, they will not tackel it.

docs/examples/explorer1_anomaly.rst
Peter230655: > This should not be necessary and points to a bug somewhere. (or misuse)

To calculate the reaction forces, the acceleratrions are needed, which are in the bottom half of the r.h.s of the eoms.
If I use ``generate_ode_function`` with ``generator=cython`` C_contiguous inputs are needed, and the solution array of solve_ivp is not C-contiguous if e.g. Radau is used. (see issue 23657 with scipy).

If ``generate_ode_function`` could covert its input arrays to be C-contiguous, all this would not be needed.
Should I raise an issue about this?

docs/examples/explorer1_anomaly.rst
moorepants: LSODA is the default integrator in PyDy's System(). It is switches to a BDF type method when stiffness is detected.

docs/examples/explorer1_anomaly.rst
moorepants: This needs to be addressed separately from this example.

The solution for this example is to use an integrator that does not have this issue or to use the default ode generator (lambdify based one).

Peter230655: > This needs to be addressed separately from this example.

I agree. Should I raise an issue about this C-contiguous matter? I think, an important matter.
> 
> The solution for this example is to use an integrator that does not have this issue or to use the default ode generator (lambdify based one).

Yes, but there are many other issues you rightly raised with this example:
- where would I quickly find a sketch and then somehow incorporate it into the .rst file?
- use the system class to avoid double book keeping; I have never used it, so I have to learn it first.

So, maybe this example has to await the next release

Peter230655:
I think, with the new keyword force_ccontiguous I can correct the stuff related to C_contiguous matters.

docs/examples/explorer1_anomaly.rst
Peter230655: > Remove if not used.

removed

docs/examples/explorer1_anomaly.rst
Peter230655: > This example really needs a figure detailing all the coordinates, forces, etc.

I added a link to a pciture of explorer1 giving some details

docs/examples/explorer1_anomaly.rst
Peter230655: > What is the purpose of this statement? To teach something? Or is is noted a bug or missing feature?

taken care of.

docs/examples/explorer1_anomaly.rst
Peter230655: > ```python
> s = System(kane)
> s.integrate()
> ```
> 
> gets everything from kane automatically (internally it populates generate_ode_function()).
> 
> generate_ode_function requires exactly what it asks for and you have to provide it.

Taken care of.

moorepants:
Please use System() for the examples.

docs/examples/explorer1_anomaly.rst
Peter230655: > This shouldn't be necessary and may point to a bug in SymPy.

It stems from the points where the antennae attach to the body of the explorer.
These points have reals speeds in the inertial frame N, but also the antennae excert forces on them.
To get these forces I add virtual speeds for these points. Whenever a points has real and virtual speeds in the same direction, the virtual speeds show up in the force vector. (not in the mass matrix, this was a miste of mine).

In a case like this, should I set up the virtual speeds in the body fixed frame of the explorer?
I just do not know.

Peter230655:
> Please use System() for the examples.

This will take me a while, I have never used it.

Peter230655:
I looked at the Carvallo-Whipple bike, I think, basically System() looks clear enough, but tow questions right now:
1. In this example, I get the auxiliary speeds and the reaction forces into the force vector, and I simply set them to zero. I do not know how to do this in System(). 
Above I explained why I think, I get them into the force vector. Should the virtual speeds be defined in the explorer fixed frame (Ae), where these points have zero actual speed? Soemthing like:
```
v1 = P0.set_vel(N, u0x * N.x + u0y * N.y + u0z * N.z)
v2 = P0.set_vel(Ae, aux0x * Ae.x + aux0y * Ae.y + aux0z * Ae.z)
P0.vel(N) = v1 + v2
```
2. For the reaction forces, the accelerations are needed. I did not see in the Cavallo-Whipple bike where to get them. Did I miss it - due to age? :-)

moorepants:
1. These examples you have proposed are quite complicated. So one suggestion is to reduce the complexity until PyDy can manage it.

2. If System() is missing a feature, then you can either add it or request it.

Peter230655:
> 1. These examples you have proposed are quite complicated. So one suggestion is to reduce the complexity until PyDy can manage it.

I can remove the reaction forces, then I should be able to re-write them for System()

> 2. If System() is missing a feature, then you can either add it or request it.

I will open an issue with pydy.

My bigger concern is whether these auxiliary speeds should ever appear in the force vector. A while back you had me convinced that they must not - not now I do not know how this would work here.
_Where would be the right place to discuss this?_

moorepants:
Forces can be functions of velocity and if those velocities are functions of auxiliary speeds, then the auxiliary speeds would appear in the forcing terms.

moorepants:
>  Where would be the right place to discuss this?

If you just want to discuss it I guess send an email. If you want to address something about the software then you can discuss it on the appropriate software repository.

Peter230655:
> Forces can be functions of velocity and if those velocities are functions of auxiliary speeds, then the auxiliary speeds would appear in the forcing terms.

But, as I understand it, reaction forces do no work, so they have no business being in the force vector.
Anyway, as you suggested, let me send an email on this subject.

Peter230655:
I changed it to using System()
I added a link to a better picture of Explorer1. I do not know how to insert the picture into the html file, like you have on your examples.
There is a link to a NASA lecture by D. Levinson, which contains a lot of technical and historical details. I do not know how to quote properly.

NB: I tried linear_solver='CRAMER', but after two hours it still had not finished. Probably the same thing as with
kane.rhs() : for larger simulations it takes very long.

Peter230655:
I used the new feature of ``evaluate_ode(.....)`` to get the values of the rhs for the reaction forces.
I set the reaction forces to zero explicitly. This may not be needed as System() does it automatically - if I understood it right.

Peter230655:
Changed as you suggested for the 3d n link pendulum.
Note; I specifically set the sys.specifieds (the reaction forces) to zero. Maybe not needed as System does this automatically. I left it so a user can se how to do it.

Peter230655:
I will push it to pst-notebooks

Issue #422, opened 2019-12-24
Move all examples to sphinx


moorepants:
jupyter_sphinx is now implmented and all examples should be moved to this format. This issue will track the progress:

- [ ] antropomorphic_robot
- [ ] babyboot
- [ ] bicycle
- [ ] body_with_arm
- [ ] chaos_pendulum
- [ ] differential_drive
- [ ] double_pendulum
- [ ] example.py
- [ ] four_bar_linkage
- [ ] inertia_of_torus
- [ ] inertia_platform
- [ ] Kane1985
- [ ] language_templates
- [ ] mae233_winter_2016_project
- [ ] mass_spring_damper
- [ ] multidof_holonomic
- [ ] multiprocessing
- [ ] npendulum
- [ ] parallel-execution
- [ ] rattleback
- [ ] README.rst
- [ ] README.txt
- [ ] rollingdisc
- [ ] rolling_disc_visualization
- [ ] scara_arm
- [ ] simple_pendulum
- [ ] sliding_ladder
- [ ] three_link_conical_pendulum
- [ ] torque-free
- [ ] tutorials
- [ ] unstable_springy_pendulum

PR #556, opened 2025-11-28
rattleback

Peter230655:
This is the rattleback from T. Kane's paper.
- the numerical integration is amazingly sensitive. Small changes in the parameters change the behaviour completely. (Maybe a reason, rattlebacks are not that prevalent in nature?)
- I never managed to get the total energy as expected, maybe due to numerical issues?
- as pydy visualization does not currently have an ellipsoid, I used a brick as substitute. I think, the basic behaviour can been seen, too.

### DIFF
--- a/docs/examples/rattleback.rst
+++ b/docs/examples/rattleback.rst
@@ -1,0 +1,552 @@
+==========
+Rattleback
+==========
+
+.. note::
+
+    You can download this example as a Python script:
+    :jupyter-download:script:`rattleback` or Jupyter notebook:
+    :jupyter-download:notebook:`rattleback`.
+
+Objectives
+==========
+
+- Show how to use ``symjit.compile_func`` as a (often) faster alternative to
+  ``sympy.lambdify`` for compiling symbolic expressions.
+- ``generate_ode_function`` from PyDy needs C - contiguous arrays as input,
+  if generator='cython' is used. Depending on the method applied in
+  ``scipy.integrate.solve_ivp``, it does not return C - contiguous arrays.
+  Show how to handle this.
+- Show a use of PyDy's visualization module to visualize the results of
+  a simulation.
+
+Description
+===========
+
+This simulation is taken from this article:
+https://www.sciencedirect.com/science/article/abs/pii/0020746282900178
+The equations of motion are set up using Kane's method from SymPy's
+``physics.mechanics`` module. The ingenious way to get the contact points of
+the rattleback with the plane is from this paper.
+
+Notes:
+======
+
+- The geometric center of the rattleback is fixed in the inertial frame N,
+  hence not fully realistic.
+- The numerical integration is quite sensitive.
+- This may be the reason the total energy is not correct.
+
+
+
+.. jupyter-execute::
+
+    import numpy as np
+    import sympy as sm
+    import symjit
+    import sympy.physics.mechanics as me
+    import matplotlib.pyplot as plt
+    from copy import deepcopy
+    from scipy.integrate import solve_ivp
+    from pydy.codegen.ode_function_generators import generate_ode_function
+    from pydy.viz.shapes import Cylinder, Sphere, Box
+    from pydy.viz.scene import Scene
+    from pydy.viz.visualization_frame import VisualizationFrame
+
+
+Set up the Equations of Motion
+------------------------------
+
+.. jupyter-execute::
+
+    N, R = sm.symbols('N, R', cls=me.ReferenceFrame)
+    O = me.Point('O')
+    t = me.dynamicsymbols._t
+    O.set_vel(N, 0)
+
+Ro is the mass center of the rattleback, S is the contact point with the plane.
+
+.. jupyter-execute::
+
+    S, Ro = sm.symbols('S, Ro', cls=me.Point)
+
+Rotation angles and speeds of R w.r.t. N
+
+.. jupyter-execute::
+
+    q1, q2, q3, u1, u2, u3 = me.dynamicsymbols('q1, q2, q3, u1, u2, u3')
+    x1, x2, x3, ux1, ux2, ux3 = me.dynamicsymbols('x1, x2, x3, ux1, ux2, ux3')
+
+symjit cannot use dynamicsymbols, so regular symbols for replacement in the
+compilation are defined
+
+.. jupyter-execute::
+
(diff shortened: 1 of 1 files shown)


Peter230655:
I may have a mistake in my description of the system.

Peter230655:
I tested the speed of the contact point after a discussion with Carlos Roithmayr. Seems correct.
I am unsure whether I think this should be merged:

On the + side.
- it rattles, and it has a nice animation.
- it shows how to adjust the size of the animation and how to rotate it
- it shows how to use symjit for compilation
- it shows how to use ``np.ascontiguousarra(..)`` 

on the - side
- the total energy does not behave as expected! This is normally a sure sign, that the description of the system is wrong. Here it may be due to numerical errors?

Peter230655:
I do not know what the failed check means. It says unknown error.

docs/examples/rattleback.rst
```diff
@@ -0,0 +1,668 @@
...
+
+    You can download this example as a Python script:
+    :jupyter-download:script:`rattleback` or Jupyter notebook:
+    :jupyter-download:notebook:`rattleback`.
+
+Objectives
+==========
+
+- Show how to use ``symjit.compile_func`` as a (often) faster alternative to
+  ``sympy.lambdify`` for compiling symbolic expressions.
```
moorepants, line 15: If you want to use symjit please use the SymjitODEFunctionGenerator, which is a feature of PyDy.

docs/examples/rattleback.rst
```diff
@@ -0,0 +1,668 @@
...
+  ``scipy.integrate.solve_ivp``, it does not return C - contiguous arrays.
+  Show how to handle this.
+- Show a use of PyDy's visualization module to visualize the results of
+  a simulation.
+
+Description
+===========
+
+This simulation is taken from this article:
+https://www.sciencedirect.com/science/article/abs/pii/0020746282900178
```
moorepants, line 27: Please make a proper restructuredtext citation.

docs/examples/rattleback.rst
```diff
@@ -0,0 +1,668 @@
...
+The equations of motion are set up using Kane's method from SymPy's
+``physics.mechanics`` module. The ingenious way to get the contact points of
+the rattleback with the plane is from this paper.
+
+Notes:
+======
+
+- The numerical integration seems very sensitive:
+
+  - method=DOP853, rtol=1.e-5, atol=1.e-6 seems the minimum to get a
```
moorepants, line 37:
```suggestion
  - ``method=DOP853, rtol=1.e-5, atol=1.e-6``` seems the minimum to get a
```

use fixed width font for such things

docs/examples/rattleback.rst
```diff
@@ -0,0 +1,668 @@
...
+
+    symbol_dict = symbol_dict_1 | symbol_dict_2
+
+    x1_sub = x11.subs(symbol_dict)
+    x2_sub = x21.subs(symbol_dict)
+    x3_sub = x31.subs(symbol_dict)
+
+    pos_kane = [x1_sub, x2_sub, x3_sub]
+    pos_jit = symjit.compile_func([qs1, qs2, qs3, us1, us2, us3], pos_kane,
+                              params=pL_pos)
```
moorepants, line 272: PyDy is not the place to demonstrate general symjit functionality. Please only use lambdify from SymPy.

docs/examples/rattleback.rst
```diff
@@ -0,0 +1,668 @@
...
+        args[-3], args[-2], args[-1] = speed_jit(*y, *args2)
+
+        # Store args for later plotting, energy, etc.
+        time_list.append(t)
+        args_list.append(deepcopy(args))
+        args = np.array(args)
+        # Ensure y is C - contiguous
+        y = np.ascontiguousarray(y)
+        rhs = rhs_gen(y, t, args)
+        return rhs
```
moorepants, line 435: The purpose of `generate_ode_function` is to generate the gradient for direct use. You shouldn't have to make a wrapper gradient. If you do, then open a bug report. We can solve that first.

Peter230655: > The purpose of `generate_ode_function` is to generate the gradient for direct use. You shouldn't have to make a wrapper gradient. If you do, then open a bug report. We can solve that first.

Not sure, whether this is a bug or a current limitation, I opened issue #561

docs/examples/rattleback.rst
Peter230655: > Please make a proper restructuredtext citation.

I must admit I do not know how to do this.

docs/examples/rattleback.rst
Peter230655: > If you want to use symjit please use the SymjitODEFunctionGenerator, which is a feature of PyDy.

I used it for speed, as it is called in every loop during numerical integration. Of course I can change to lambdify.

docs/examples/rattleback.rst
Peter230655: > PyDy is not the place to demonstrate general symjit functionality. Please only use lambdify from SymPy.

I did it for speed, it is called in every loop of the numerical integration. Of course I can change to lasmbdify

Peter230655:
I will close this PR for those reasons:
1. The total energy is not exactly constant. This may be due to mis-modeling or to numerical errors, I don't know. We should not have simulations here, where there is doubt about them being correct.
2. I need to numerically calculate the location of the contact point of the rattle-back with the plane at every step of the integration. At present, System does not allow this (as I understand it anyway), so I need a wrapper - defeating the present set up of System.

I may push it to pst-notebooks some time, as there is more leeway for both points.

moorepants:
Note that there are examples of the rattleback already in pydy: https://github.com/pydy/pydy/tree/master/examples/rattleback

I have not read them in detail in many years, but if the do not require numerical calculations at each time step, one or more of these could be transformed into a documentation example (as per #422).

Peter230655:
I saw them before I started my rattleback, but since they did not seem to be solved (numerically integrated I started my own version. (I have been interested in this for several years...)

I checked my simulation again. I may get by without numerically solving during integration (I paid no attention then to this point at the time).
If you are interested in having this simulation, I will try.

moorepants:
Those examples do generate numeric codes.

Peter230655:
I must have missed this.
So, another rattleback obviously not needed.

PR #558, opened 2026-04-20
Add Baumgarte's stabilization to bicycle model.

moorepants:
This will need some tidying but I read Baumgarte's original paper and realized it was rather easy to include his stabilization technique for constraints. You simply add a spring and damper to the acceleration level constraints for any holonomic constraints in which the spring displacement is the residual of the holonomic constraint.

If I simulate the bicycle example for 20 seconds with no stabilization this is the holonomic constraint residual:

image

And with alpha = 10:

image

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -27,6 +27,6 @@ jobs:
           python-version: ${{ matrix.python-version }}
           channels: conda-forge
       - name: Install basic dependencies
-        run: conda install -q -y cython jupyter_sphinx "jupyter_core<5" matplotlib numpy numpydoc pythreejs setuptools "sphinx<7" symjit sympy scipy
+        run: conda install -q -y "cython=0.29.*" "jupyter_sphinx=0.3.*" "matplotlib-base=3.5.*" "numpy=1.21.*" "numpydoc=1.2.*" pythreejs "setuptools=44.1.*" "sphinx=4.3.*" symjit "sympy=1.9.*" "scipy=1.8.*"
       - name: Test building documentation
         run: cd docs && make clean && make html && cd ..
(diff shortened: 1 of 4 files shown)


Peter230655:
Would $\alpha^2$ correspond to the spring and $2 \cdot \alpha$ to the dampening?
So, in a way you pull the point which should meet the holonomic constraint back to the constraint if it wants to leave it?

moorepants:
Yes, correct.

Peter230655:
Would this dampening affect the energy balance of the system, or would it be 'perpendicular' to actual motions?
(I am not good enough with Kane to see this)

moorepants:
It only adds a term that is equal to zero. So it should not affect energy, at least to a tolerance level.

Peter230655:
> It only adds a term that is equal to zero. So it should not affect energy, at least to a tolerance level.

It is only zero if the holonomic constraint is kept (?)
You may not recall, but a long time ago I did some simulation where some point had to stay fixed in space.
It did not work right, so I connected a very strong spring between the point and the location where it should be.
. You agreed this was fine.
I did not add any damping, though. The spring energy was neglegible even though the spring was VERY strong. I guess here the effect must be even much less.

moorepants:
Close in favor of #559 (I screwed up the commits locally).

Peter230655:
I tried Baumgarte's method on my n link closed pendulum, where the second attachment point is a holonomic constraint.
So far it does not seem to help much, with certain (small) values of $\alpha$ I get a small improvement, with most values a drastic deterioration. Maybe I am doing something wrong.

moorepants:
There are many papers on the topic: https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=baumgarte+stabilization&btnG=&oq=baumgar

PR #572, opened 2026-04-28
Support passing in values to evaluate_ode(), add evaluate_constraints(), add set_dependent_initial_conditions()

### LINKED ISSUE #570: Add functionality to System() to return the state derivatives
First we need to support this:

```python
x = n.ones(n)
system.evaluate_ode(x=x, t=t, r=r, p=p)
```

i.e. let the user override the values used.

And then you can also support managing an entire trajectory:

```python
x  # shape(m, n), m: # time steps, n: # states
times # shame(m,)
xdot = system.evaluate_ode(x=x_traj, t=times)
xdot  # shape(m, n)
```

### LINKED ISSUE #571: Add methods for evaluating the constraints to System()
This can be similar to `evaluate_ode()`.

```python
system.evaluate_constraints(x=None, t=None, r=None, p=None)
system.evaluate_holonomic(x=None, t=None, r=None, p=None)
system.evaluate_nonholonomic(x=None, t=None, r=None, p=None)
```

Default is to use the initial conditions and other values set on System. You can override as suggested in #570.

moorepants:
Fixes #570
Fixes #571 

TODO:

- [x] add unit tests for the new methods


Peter230655:
As soon as it is merged, I will test it with my two simulations.

pydy/system.py
```diff
@@ -456,6 +457,175 @@ def _kwargs_for_gen_ode_func(self):
...
+        else:
+            x = np.asarray(x)
+        p = np.array([self.constants[pi] for pi in self.constants_symbols])
+        if len(x.shape) == 2:
+            x = x.T
+            return self._eval_nonholonomic(x, p).squeeze().T
+        else:
+            return self._eval_nonholonomic(x, p).squeeze()
+
+    def set_dependent_initial_conditions(self, use_jacobian=False, tol=1e-12):
```
moorepants: Right now you have to set the independent states as named in the KanesMethod call and it solves for the dependent. This works, but I could allow you to select any p + n states to declare and the other m + M are solved for. That would be more flexible.

Peter230655:
I agree, more flexible. But I cannot right now think of an example where this would be needed.
If I have m dependent coordinates and m dependent speeds, I need to get them for consistent initial values, not more and not less?

Peter230655:
I my experience root is definitely better than fsolve.

moorepants:
> But I cannot right now think of an example where this would be needed.

I just added one.

> I my experience root is definitely better than fsolve.

There is no difference in the two if the default method is used.

Peter230655:
> > But I cannot right now think of an example where this would be needed.
> 
> I just added one.
> 
> > I my experience root is definitely better than fsolve.
> 
> There is no difference in the two if the default method is used.

Of course my statement onlyx based on my experience. I did not look at their inner workings.

moorepants:
The oldest build fails if the Jacobian is passed to root(). This seems to point to a possible bug in scipy 1.11.4.

moorepants:
This issues seems to be similar to the one I'm hitting and only fails with the exact Jacobian (which we derive here): https://github.com/scipy/scipy/issues/16938

EDIT: also similar https://github.com/scipy/scipy/issues/16789

moorepants:
It seems that it does find the correct solution even though root() reports a failure (same as explained in the scipy issue).

pydy/tests/test_system.py
```diff
@@ -57,7 +57,7 @@ def test_init(self):
         assert sys.specifieds == dict()
         assert sys.initial_conditions == dict()
         assert sys.constants == dict()
-        assert sys.times == list()
+        np.testing.assert_allclose(sys.times, np.array([]))
```
moorepants, line 60: This is a backwards incompatible change.

moorepants: It is fairly unlikely that anyone relied on this default value for anything.

moorepants:
I think I'm finally happy with this.

PR #576, opened 2026-05-04
Rattleback

Peter230655:
Changed it to using System()
I am finally convinced it is right:
- energy drops as expected
- the contact point moves as expected
- the rattleback does change its main direction of rotation

I did not push this with a desire to have it merged (although I think it is a nice simulation), but so that you can see it.


Peter230655:
I made the changes suggested with 3d - pendulum.
I added the new features to calculate the violations of the configuration constraints and of the velocity constraints.

I think (of course I am not completely neutral...) that this is an nice example:
- it works. I am finally satisfied it is correct
- it uses all the new additions to pydy
- it is reasonably simple, there are no "hacks" used anywhere

BUT the big drawback: it takes time to run. The line
``sys = System(kane, ode_solver=ode_solver)`` takes 75 sec to run. the line
``resultat = sys.integrate(method='DOP853', atol=1.e-8, rtol=1.e-8)``  takes 55 sec. If I lower atol / rtol the results will be wrong, e.g. the energy does not decrease.

If we cannot use it here, I think, we wait until the next release of pydy so I can push it to pst-notebooks.
(of course, I'd prefer for it to be here - but the total running is over 140 sec.)

Peter230655:
I looked a bit closer at the rattleback in pydy/examples (not in pydy/docs/examples)
- they derived the eons from first principles, I used sympy.physics.mechanics Kane's method.
- They did two surfaces, an ellipsoid and an parabolic shape, here only the ellipsoid.
- There is much more theory in some text file, I simply followed the paper cited here and by them, too.
- I could not find any numerical calculations and resulting graphs, but maybe they are there and I simply did not find them.
- Here, of course, all features of pydy are used, which I assume were not available 11 years ago.

In summary, I would not know how to make this into a simple example. I think, too elaborate to fit in here where we want simple examples showing how to use the features of pydy. 
The example here runs too long, 140 sec! Naturally I do not know whether the eoms they came up with would run much faster.

Peter230655:
I will move this to pst-notebook. As we discussed it is too complicated to be suitable here, and also runs too long.

PR #583, opened 2026-05-13
Add outputs calculation to System() and enable calculation noncontributing forces

### LINKED ISSUE #575: Speed up System()'s search for the constants and specifieds
On instantiation of System() the full mass matrix and forcing vector are searched. This is unnecessary as these could be extracted from the list of loads and bodies. If we searched the force vectors and the body velocities instead all the symbols can be found. There is a sligth negative as some constants and specifieds may not show up in the full equations of motion even though they are present in the loads and bodies.

### LINKED ISSUE #577: If you add an ode_solver to System after instantiation and it is a CythonGenerator, the force_c_contiguous flag is not set

### LINKED ISSUE #578: Support for calculating reaction forces
Kane's method allows to calculate reaction forces via the ausiliary_eqs.
System() does not support this.
Of course there are ways around this, but it would be nice to have somthing like
``evaluate_reaction_force(x=solution)`` returning the reaction forces.

### LINKED ISSUE #587: Support outputs in System()
Copied over from #136:

And idea on a generic API:

```python
fs, V, r, x, y = dynamicsymbol('f_s, x, y')
k, l, g, m = symbol('k')

s = System(
    method,
    outputs={
        fs: k*x,  # 1: spring force
        V: m*g*y,  # 2: potential energy
        r:  x**2 + y**2 - l,  # 3: constraint residual
        Matrix([TA, TB]): Matrix([m/2*va**2, m/2*vb**2]),  # 4: two kinetic energy equations for body A and body B
        Matrix([T1, T2]): Matrix([v.diff()/m - k*x + T2 - T1, va.diff()/m - k*x /m+ vb.diff()/m - k*x/m - T2]),  # 5: noncontributing forces, that need to be solved (linear solve)
        a_sq: va.diff()**2 + vb.diff()**2,  # 6: nonlinear in the generalized accelerations 
    }
)
```

For 1, 2, 3, and 4, these could be included as an extra column matrix of expressions to evaluate, either by making a distinct output `xdot, y = rhs()` or making dummy states that gives `vstack((xdot, y)) = rhs()`.

For 5, we could check that T1, T2, va', vb' are present in the provided equations and also if they are linear, then augment the mass matrix for the simultaneous linear solve. This works for passing in Kane's auxiliary equations but not Lagrange multipliers for constraint forces. That would still need a special case as it modifies the dynamical differential equations.

But method will always be aware of whether it has auxiliary equations and/or constraints and the augmentation can happen internally.

For 6, this is more challenging because we don't, in general, do the linear solve symbolically and thus don't have access to the udots for computing outputs. You basically need a separate function that ingests the udots. So, 6 can't really be wrapped into the code generation to take advantage.

Thinking about this more generally makes me question whether I should have prematurely added `evaluate_constraints()`, as this can also just be treated as an "output equation".

This style of APi fits with what we have for specifieds.

moorepants:
This will fix #578 and is generalizable to #127. It uses the approach described in #127.

Fixes #587
Fixes #578
Fixes #577
Fixes #575

# Explanation

You can now add output expressions to evaluate alongside the state. For example:

```python
outputs = {
    k(t): expr1,
    (w(t), h(t)): (expr2, expr3),
    (f1(t), f2(t)): (a*f1 + b*f2 + c*u1' + d*u2' + expr4, e*f1 + f*f2 + g*u1' + h*u2' + expr5), 
}
s = System(..., outputs=outputs)
```
a, b, c, d, e, f, g, h, expr1, expr2, expr3, expr4, expr5 are all, at most, functions of (q, u, r, p, t).

The (f1, f2) includes f1 and f2 in the expressions which represent a system of equations linear in f1 and f2 and possibly the udots. Nonlinear functions of f1, f2, u1', and u2' are not supported.

Once the outputs are added to System, they can be evaluated like so:

```python
y = s.evaluate_outputs()
```

where y is an ndarray with values [k, w, h, f1, f2].

In addition, if KanesMethod has constraints these are automatically added as outputs and if it has auxiliary equations these are optionally added as outputs. If constraints are present they can be evaluated by:

```python
fh = s.evaluate_config_constraints()  # shape(M,)
fn = s.evaluate_motion_constraints()  # shape(m + M,) [KanesMethod requires passing in the derivative of fh]
fc = s.evaluate_constraints()  # shape(m + 2*M,)
```

If there are constraints, these are added to the outputs, so:
```python
assert fc == s.evaluate_outputs()
```
passes if no other user supplied outputs were present.

If there are auxiliary equations, then these are added as linear outputs and the results `lam` are accessed with:

```python
x = s.integrate()
res = s.evaluate_ode(x=x)
xd = res[:len(s.num_coordinates + s.num_speeds)]
lam = res[len(s.num_coordinates + s.num_speeds):]
```

All of this is designed so that the `ODEFunctionGenerator` creates a single optimized function to evaluate all state derivatives and all outputs simultaneously. It has a negative consequence that if you only want to evaluate a single output, you must evaluate everything (state derivatives and output equations). But the positive is that these things can be very efficient and the user has access to calling `s.evaluate_ode_function()` directly for optimal execution speed.

## New API in this PR

- .auxiliaries: list of internally generated auxiliary states which are the time integrals of the noncontributing forces or any linear system variables in .outputs
- .noncontributing_forces: list of undefined functions of time that represent the noncontributing forces exposed via the auxiliary generalized speeds in Kane's method.
- .constraints: column matrix of sympy expressions that are functions of q, u,[ r,] p representing configuration and motions constraints, the motion constraints can be time derivatives of the configuration constraints (so redundant)
- .num_constraints: integer
- .num_config_constraints: integer
- .num_motion_constraints: integer
- .num_outputs: integer
- .outputs: dictionary mapping undefined functions of time or tuples thereof to expressions to evaluate. it support expressions that are linear in the functions and/or udots.
- .outputs_symbols: list of undefined functions of time for each output in .outputs
- .evaluate_outputs(): returns the numerical values of all outputs in .output_symbols (in same order), works like .evaluate_ode()

# API Options

## Return Auxiliary States?

Auxiliary states are created for any output that was linear in the generalized
accelerations.

```pycon
>>> outputs = {
...    (f1, f2): ((f1 + f2 - 10), (2*f1 + 4*f2 - 20))
... }
>>> s = System(..., outputs=outputs)
>>> s.states
x(t), v(t), ∫ f1 dt, ∫ f2 dt
>>> xd = s.evaluate_ode()
>>> xd.shape
(4,)
```

We could hide these dummy states so that you instead see:

```pycon
>>> s.states
x(t), v(t)
>>> s._actual_states
x(t), v(t), ∫ f1 dt, ∫ f2 dt
>>> xd = s.evaluate_ode()
>>> xd.shape
(2,)
```

but this may break the ability to do:

```pycon
x = s.integrate()
xd = s.evaluate_ode(x=x)
```

The necessary dummy state values would be missing to evaluate x'.

## Name of noncontributing/constraint load symbols

We need to provide the dynamic symbols that correspond to the unknown noncontributing forces. Ideally this would map to the force or torque measure numbers present in the auxiliary equations in Kane's method as well as the constraint force or torque measure numbers exposed as Lagrange multipliers that are associated with any explicit holonomic or nonholonomic constraints.

The current name is `noncontributing_symbols`.

Attributes that are list or sets of symbols:

- coordinates (list of [generalized] coordinates)
- speeds (list of [generalized] speeds)
- auxiliaries (time integral of the Lagrange Multipliers/Noncontributing Forces)
- constants_symbols (associated with constants dict, can optionally be set only on System instantiation)
- specifieds_symbols (associated with specifieds dict, can optionally be set only on System instantiation)
- outputs_symbols (associated with outputs dict)
- lagrange_multipliers (associated with LagrangesMethod.constraints, can optionally be set only on System instantiation), should _symbols be appended?
- noncontributing_forces (associated with LagrangesMethod.constraints, can optionally be set only on System instantiation), should _symbols be appended?

Maybe it is fine to not append `_symbols` because they are really only associated with internal proprieties of Kanes/Lagranges/TMT/Method.

Symbols for the noncontributing forces are associated with the kane.auxiliary_eqs placed in the outputs dict. It is also associated with the auxiliaries. Should the names of auxiliaries and the symbols for the noncontributing forces have more similarity.

- Lagrange/TMT: add Lagrange multipliers that represent the constraint force measure numbers
- Kane: add noncontributing force measure numbers

Options for naming the list of symbols:

- .lagrange_multipliers_symbols: I don't think that Kane's approach for noncontributing forces is the Lagrange multiplier approach (see: https://en.wikipedia.org/wiki/Lagrange_multiplier), so this name may only work for Lagrange/TMTMethods.
- .force/load_measure_numbers_symbols: these are the measure numbers of force vectors for all methods, whether we use "force" or "load" is a debate. "force" sounds right but "load"= force or torque, whereas saying "force" assumes we also include torque. Kane uses "force" in his writing and most other dynamcis text say "force", like "generalized forces" in Lagrange explanations.
- .constraint_forces/loads_symbols
- .noncontributing_forces/loads_symbols
- We can have duplicate names that are redundant in the future, e.g. noncontributing_forces for Kane's and lagrange_multipliers or constraint_forces for Lagrange/TMT.

# Considerations for Future Features and API

## Lagrange Multipliers

Lagrange multipliers can be used to include constraint forces. SymPy's
LagrangesMethod outputs this form for constrained systems. I'm also working on
a TMTMethod that will also output the same result.

```
[Md C.T][u' ] = [Fd]
[C  0  ][lam]   [Fc]
```

I do not see a clean way of letting the user provide this via System.outputs
because it involves augmenting the dynamical differential equations (unlike
Kane's method of exposing noncontributing forces).

This can be easily managed internally by System in `_kwargs_for_gen_ode_func`,
with something roughly like:

```python
Md = self.eom_method.mass_matrix
C = self.eom_method.jacobian_of_constraints
CZ = C.row_join(sm.zeros(*C.shape))
mass_matrix = Md.row_join(C.T).col_join(CZ)
```

## Support Output Names

The only functionality this really enables is printing a useful list of names
in the autogenerated rhs docstring.

current:

```pycon
>>> outputs = sm.Matrix([expr1, expr2])
>>> rhs = generate_ode_function(..., outputs=outputs)
>>> help(rhs)
...
Returns
=======
dx : ndarray, shape(8,)
    The derivative of the state vector.
y : ndarray, shape(2,)
    Values of the provided outputs.
        - y1(t)
        - y2(t)
```

option 1:

```python
outputs = {(a(t), b(t)): sm.Matrix([expr1, expr2])}
rhs = generate_ode_function(..., outputs=outputs)
```

option 2:

```python
outputs = {a(t): expr1, b(t): expr2}
rhs = generate_ode_function(..., outputs=outputs)
```

option 3:

```python
names = (a(t), b(t))
outputs = sm.Matrix([expr1, expr2])
rhs = generate_ode_function(..., outputs=outputs,
                            outputs_names=names)
```

```pycon
>>> help(rhs)
...
Returns
=======
dx : ndarray, shape(8,)
    The derivative of the state vector.
y : ndarray, shape(2,)
    Values of the provided outputs.
        - a(t)
        - b(t)
```

## Support Multiple Output Arrays

```python
outputs=(
    sm.Matrix([expr1, expr2]),
    sm.Matrix([expr3, expr4]),
    sm.Matrix([expr5]),
)
rhs = generate_ode_function(..., outputs=outputs)
```

```pycon
>>> help(rhs)
...
Returns
=======
dx : ndarray, shape(8,)
    The derivative of the state vector.
y1 : ndarray, shape(2,)
    Values of the provided outputs.
        - y11(t)
        - y12(t)
y2 : ndarray, shape(2,)
    Values of the provided outputs.
        - y21(t)
        - y22(t)
y3 : ndarray, shape(2,)
    Values of the provided outputs.
        - y31(t)
        - y32(t)
```

## Support Output Matrices, e.g. Jacobian

```python
outputs = sm.Matrix([[expr1, expr2],
                     [expr3, expr4]])
rhs = generate_ode_function(..., outputs=outputs)
```

```pycon
>>> help(rhs)
...
Returns
=======
dx : ndarray, shape(8,)
    The derivative of the state vector.
y : ndarray, shape(2,2)
    Values of the provided outputs.
        | y11(t), y12(t) |
        | y21(t), y22(t) |
```

What about names in this case? I guess if a single name is given for a Matrix,
then:

```python
outputs = {m(t): sm.Matrix([[expr1, expr2],
                            [expr3, expr4]])}
rhs = generate_ode_function(..., outputs=outputs)
```

```pycon
>>> help(rhs)
...
Returns
=======
dx : ndarray, shape(8,)
    The derivative of the state vector.
m(t) : ndarray, shape(2,2)
    Values of the provided outputs.
        | m11(t), m12(t) |
        | m21(t), m22(t) |
```

otherwise you can name each element:

```python
outputs = {((a(t), b(t)),
            (c(t), d(t))):
           sm.Matrix([[expr1, expr2],
                      [expr3, expr4]])}
rhs = generate_ode_function(..., outputs=outputs)
```

```pycon
>>> help(rhs)
...
Returns
=======
dx : ndarray, shape(8,)
    The derivative of the state vector.
y : ndarray, shape(2,2)
    Values of the provided outputs.
        | a(t), b(t) |
        | c(t), d(t) |
```

## Support outputs that are nonlinear functions of xdot

```python
outputs = {
    a(t): u3.diff(t),  # simply return a generalized acceleration (note that a(t): a(t) - u3.diff(t) currently works)
    f(t): sqrt(u1.diff(t)**2 + u2.diff(t)**2),  # nonlinear function of generalized accelerations
    g(t): cos(g(t))*sin(u1.diff()),  # would require a root() solve internally to find g(t)
}
rhs = generate_ode_function(..., outputs=outputs)
```

# TODO

- [x] Should the dummy states (impulse of constraint forces) be hidden from the user?, i.e. only present in the ode_evaluation_function
- [x] Sort out #587 so that this can make use of the more general feature.
- [x] Make sure constraints are automatically passed to outputs.
- [x] Fix order of output of evalaute_outputs().
- [x] Can the API support future generation of the Jacobian? (at least not be limiting in implementing this)
- [x] Cand the API support generating output groups (for example the Jacobian above not being mixed with other outputs).
- [x] Can the API support outputs that are nonlinear functions of the generalized accelerations?
- [x] Can the API support Lagrange multiplier style augmented equations of motion?
- [x] Decide on what attributes/methods should be public and the best names.
- [x] Should the user pass in names for the constraint residuals?
  - they can access the generated dummy symbols in system.outputs if needed, we can leave this be 
- [x] <strike>Maybe make the code generator outputs have names so the docstring shows something more than y1 through yn.</strike> open issue for this
- [x] Figure out the default value of sys.times
- [x] Decide on name of `constraint_loads` kwarg.
  - I plan to use `System(..., noncontributing_forces=[F1, F2], lagrange_multipliers=[lam1, lam2])` and they will just be redundant. I suppose you could technically do both together, so maybe this leaves that open.
- [x] <strike>evaluate_constraints should just call evaluate_outputs</strike>, open issue for this
- [x] add tests for providing `constants_symbols` and `specifieds_symbols`
- [x] Search output equations for constants and specifieds.


pydy/codegen/tests/test_ode_function_generator.py
```diff
@@ -415,10 +415,14 @@ def setup_method(self):
...
         self.rhs = self.sys.eom_method.rhs()
+        kinetic_energy = []
+        for b in self.sys.eom_method.bodies:
+            kinetic_energy.append(b.kinetic_energy())
         self.generator = ODEFunctionGenerator(self.rhs,
                                               self.sys.coordinates,
                                               self.sys.speeds,
-                                              self.sys.constants_symbols)
+                                              self.sys.constants_symbols,
+                                              outputs=kinetic_energy)
```
moorepants: Remove this change, accidentally merged.

pydy/system.py
```diff
@@ -97,7 +97,8 @@ class System(object):
 
     """
     def __init__(self, eom_method, constants=None, specifieds=None,
-                 ode_solver=None, initial_conditions=None, times=None):
+                 ode_solver=None, initial_conditions=None, times=None,
+                 constraint_loads=None):
```
moorepants: options for the name:

- constraint_forces
- constraint_loads
- noncontributing_forces
- noncontributing_loads
- reaction_forces
- reaction_loads

Peter230655: I like noncontributing_forces best - but I am not a native English speaker.

moorepants: I eventually want this to work with LagrangesMethod and new TMTMethod class. And noncontributing doesn't align so well with those methods' terminology.

moorepants: I've changed it to `noncontributing_symbols`. The `_symbols` makes it consistent with other iterables of symbols `constants_symbols`, `specifieds_symbols`, `outputs_symbols`. In Lagrange's method you add Lagrange multipliers and typically say these represent "constraint forces". In Kane's method you add auxiliary speeds to expose noncontributing forces. The constraint forces are also noncontributing, so the work can apply to both. It is also true that the noncontributing forces are constraint forces but are not associated exclusively with the constraint forces due to holonomic or nonholonomic constraints. I can always add a duplicate kwarg called "constraint_forces_symbols" or something if I want the name to better align.

We also have `constraints` which returns the column matrix of constraint expressions. Having `constraints_symbols` can be confusing because we also create names for the residuals of the constraints internally.

moorepants:
@Peter230655 this is very close to being done. I'm still working out some details and have to make some API design decisions. If you have any opinions on what I have now, this is a good time to let me know.

Peter230655:
> @Peter230655 this is very close to being done. I'm still working out some details and have to make some API design decisions. If you have any opinions on what I have now, this is a good time to let me know.

As soon as you push it in the developmental version I will play aorund with it.
Then I can see how easy it is to use. (I afraid to use your branch directly - likely I mess up my environments....)

moorepants:
I'm not going to merge this until I think it is done. The point of the PR is to have a way to review code without merging it.

moorepants:
You also do not have to use it to review it. You can read everything I have written, for example.

Peter230655:
> You also do not have to use it to review it. You can read everything I have written, for example.

Dumb question: can the dependent speeds not be calculated from the speed constraints and the independent speeds?
image

moorepants:
> Dumb question: can the dependent speeds not be calculated from the speed constraints and the independent speeds?

That is what the new function `set_dependent_initial_conditions()` does. See the docstring for an explanation.

moorepants:
My recommendation for reviewing this is to first read the updates to System's documentation:

https://pydy--583.org.readthedocs.build/en/583/system.html

Then you can look at the new unit tests that show exhaustively how to use the new features with a double pendulum system.

Peter230655:
> My recommendation for reviewing this is to first read the updates to System's documentation:
> 
> https://pydy--583.org.readthedocs.build/en/583/system.html
> 
> Then you can look at the new unit tests that show exhaustively how to use the new features with a double pendulum system.

Thanks!

Peter230655:
I think better. In sympy.physics.mechanics you call them constraint forces, so this term is closer to symoy.

moorepants:
Where do you see something called "constraint forces"? This page doesn't seem to mention that: https://docs.sympy.org/dev/modules/physics/mechanics/api/kane_lagrange.html The documentation there is a bit vague on what the noncontributing forces are and how to access them in KanesMethod and it only mentions Lagrange Multipliers in the LagrangesMethod part.

moorepants:
I found this example which calls them "constraint forces": https://docs.sympy.org/dev/tutorials/physics/mechanics/rollingdisc_example_kane_constraints.html

moorepants:
I've settled on this (above):

> I plan to use `System(..., noncontributing_forces=[F1, F2], lagrange_multipliers=[lam1, lam2])` and they will just be redundant. I suppose you could technically do both together, so maybe this leaves that open.

Peter230655:
> I found this example which calls them "constraint forces": https://docs.sympy.org/dev/tutorials/physics/mechanics/rollingdisc_example_kane_constraints.html

This is the example I meant.

moorepants:
It says "we are bringing the non-contributing (constraint) forces", so it is using both terms. But this isn't API language, so it is easily changed in SymPy. I don't think it should be the decided vote.

Peter230655:
What I meant was: I like force better than symbol (even though some real number is normally not a force)

moorepants:
The reason to append `_symbols` would be consistency with the existing API. Please see the explanations above regarding this.

moorepants:
@Peter230655 this is done from my perspective, unless you have any feedback.

Peter230655:
I looked at https://pydy--583.org.readthedocs.build/en/583/system.html as you recommended at it all looked clear to me.
If things come up when I play around with it, I will report as usual.

PR #585, opened 2026-05-15
Add support for output equations to ODEFunctionGenerator

### LINKED ISSUE #136: Support output equations in the code generators.

### LINKED ISSUE #338: Improve documentation of generate_ode_function
There are 3 different ways to call it. We should a simple for each one with and without optional arguments.

moorepants:
Fixes #136
Fixes #338

A new kwarg that accepts a column matrix of expressions that are functions of the coordinates, speeds, constants, and specifieds enables code generation of these output expressions alongside the expressions for evaluating the state derivatives. These output computations share common subexpressions and thus evaluate efficiently when both the state derivatives and the outputs are required.

TODO

- [x] Update docs page with examples of using outputs.
- [x] Consider supporting output groups, i.e. `xdot, y1, y2, y3 = rhs()`.
- [x] Consider letting the user define symbols for names of the outputs.
- [x] Add test for rhs docstring for outputs.

### DIFF
--- a/docs/codegen.rst
+++ b/docs/codegen.rst
@@ -87,8 +87,7 @@ classes:
 
 .. code:: pycon
 
-   >>> from numpy import array
-   >>> from pydy.models import multi_mass_spring_damper
+   >>> import numpy as np
    >>> from pydy.codegen.ode_function_generators import generate_ode_function
    >>> sys = multi_mass_spring_damper()
    >>> sym_rhs = sys.eom_method.rhs()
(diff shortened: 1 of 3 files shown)


moorepants:
Both adding output groups and names can be added in the future in backwards compatible ways. So I think I'll skip that. I don't think I'll support outputs that are functions of the mass matrix linear solve either, as that isn't easy to do cleanly in a single code generation.

PR #606, opened 2026-06-02
Adds support to the code generators for equations of motion that include time explicitly.

### LINKED ISSUE #133: Support explicit functions of time in the ODE function generators.
Currently, explicit functions of time in the EoMs will not work with the ode function generators. For now, you must define any explicit functions of time as generic specifieds and do the evaluation in a custom function that is passed into the rhs function. This should be required and explicit functions of time should just work.

### LINKED ISSUE #396: The System class should automatically take care of dynamicsymbols._t
Consider this example:
```
m, k, b, g=sm.symbols('m k b g', real=True)
position, speed = me.dynamicsymbols('position speed')
positiond, speedd = me.dynamicsymbols('position speed', 1)
o = me.dynamicsymbols('o')
force = o*sm.sin(me.dynamicsymbols._t)
frame_ceiling=me.ReferenceFrame('ceiling')
point_origin=me.Point('origin')
point_origin.set_vel(frame_ceiling, 0)
particle_block=me.Particle('block', me.Point('block_pt'), sm.Symbol('m'))
particle_block.point.set_pos(point_origin, position*frame_ceiling.x)
particle_block.mass = m
particle_block.point.set_vel(frame_ceiling, speed*frame_ceiling.x)
force_magnitude = m*g-k*position-b*speed+force
force_block = (force_magnitude*frame_ceiling.x).subs({positiond:speed})
kd_eqs = [positiond - speed]
forceList = [(particle_block.point,(force_magnitude*frame_ceiling.x).subs({positiond:speed}))]
kane = me.KanesMethod(frame_ceiling, q_ind=[position], u_ind=[speed], kd_eqs = kd_eqs)
fr, frstar = kane.kanes_equations([particle_block], forceList)
zero = fr+frstar
from pydy.system import System
sys = System(kane, constants = {m:1.0, k:1.0, b:0.2, g:9.8},
specifieds={me.dynamicsymbols._t:lambda x, t: t, o:2},
initial_conditions={position:0.1, speed:-1*1.0},
times = np.linspace(0.0, 10.0, 10.0/0.01))

y=sys.integrate()
```

Here we have `force = o*sm.sin(me.dynamicsymbols._t)`. In the simulation code using the System class, passing `dynamicsymbols._t` as a specified throws an error:

```
  File "C:\Users\pappu\Desktop\Diving_Into_SymPy\autolev-test-examples\pydy-example-repo\mass_spring_damper.py", line 29, in <module>
    times = np.linspace(0.0, 10.0, 10.0/0.01))
  File "C:\Users\pappu\Desktop\virtual_env\env\lib\site-packages\pydy\system.py", line 116, in __init__
    self.specifieds = specifieds
  File "C:\Users\pappu\Desktop\virtual_env\env\lib\site-packages\pydy\system.py", line 249, in specifieds
    self._check_specifieds(specifieds)
  File "C:\Users\pappu\Desktop\virtual_env\env\lib\site-packages\pydy\system.py", line 306, in _check_specifieds
    self._assert_is_specified_symbol(k, symbols)
  File "C:\Users\pappu\Desktop\virtual_env\env\lib\site-packages\pydy\system.py", line 261, in _assert_is_specified_symbol
    symbol))
ValueError: Symbol t is not a 'specified' symbol.
```
Instead if `dynamicsymbols('t')` is used (which is just a regular dynamicsymbol) in both the places mentioned above, the code works but this isn't an ideal change.
One could just pass in `force` itself as a specified and this would work just fine but it would be nice if cases like these work as well.

A convenient feature that can be added to fix this is to let the System class directly substitute time stamps in any stray `dynamicsymbols._t` occurrences in the kane's equations.

moorepants:
Fixes #133
Fixes #396

### DIFF
--- a/pydy/codegen/ode_function_generators.py
+++ b/pydy/codegen/ode_function_generators.py
@@ -428,8 +428,6 @@ y : ndarray, shape({num_outputs},)
         """Returns an ndarray containing the numerical values of the
         constants in the correct order. If the constants are already an
         array, that array is returned."""
-        print(args)
-
         if self.constants:
             p = args[-1]
         else:
--- a/pydy/codegen/ode_function_generators.py
+++ b/pydy/codegen/ode_function_generators.py
@@ -614,7 +614,7 @@ y : ndarray, shape({num_outputs},)
 
     def _create_base_rhs_function(self):
         """Sets the self._base_rhs function. This function accepts arguments
-        in this form: (q, u, p) or (q, u, r, p)."""
+        in this form: (q, u, t, p) or (q, u, t, r, p)."""
 
         if self.system_type == 'full rhs':
 
--- a/pydy/codegen/ode_function_generators.py
+++ b/pydy/codegen/ode_function_generators.py
@@ -547,13 +547,15 @@ y : ndarray, shape({num_outputs},)
                 q = args[x_idx][:self.num_coordinates]
                 u = args[x_idx][self.num_coordinates:]
 
+                t = np.array([args[1]])
+
                 if self.constants:
-                    return self._base_rhs(q, u, *args[2:])
+                    return self._base_rhs(q, u, *args[2:], t)
                 else:
                     if self.specifieds is None:
-                        return self._base_rhs(q, u, [])
+                        return self._base_rhs(q, u, [], t)
                     else:
-                        return self._base_rhs(q, u, *(args[2:3] + ([],)))
+                        return self._base_rhs(q, u, *(args[2:3] + ([],)), t)
 
             rhs.__doc__ = self._generate_rhs_docstring()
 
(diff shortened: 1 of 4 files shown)


Peter230655:
Very nice!!
I want to change my simulation to this, so no hack needed, but I am unsure how to do it.

moorepants:
There is nothing new to do. It just works.

Peter230655:
Definitely the best way! :-=)

Peter230655:
Works!
Would we like a simulation in pydy/docs/examples which shows:
- no need to worry about explicit time in the eoms
- no need to worry about noncontributing forces if the point has both real and virtual speeds ?

Based on my experience with the old way, I feel it might be helpful to a user of pydy.System - but you know better.
(otherwise I can push it to pst-notebooks, where I would show the difference between the use of pydy.System and the old way)

moorepants:
I don't think we need new examples for such things. There is no reason the first should not work, as SymPy models can have time as an explicit symbol. And the second seems like a bug.

Peter230655:
In principle, YES, you are right! But then again
- pydy.System could not handle explicit time dependence until maybe 3 hours ago. And YES, sympy Kane's Method can have time as an explicit symbol, but integrating numerically the old way still requires some 'detours.
- the 'bug' in sympy.phyisics.mechanics, Kane's Method has been around for many years, you recall the discussion in sympy's 29002

So, I feel that 'advertizing' these advantages of the most current features of System with a simulation on pst_notebooks does not hurt,

moorepants:
You can add anything to pst notebooks. Feel free. But we should not add entirely new examples here to show something we can show with one sentence in the docstrings or with a few lines added to an existing example.

PR #278, opened 2015-10-17
Add regression test for n-link pendulum model

oliverlee:
This commit resolves https://github.com/pydy/pydy/issues/108.

PR #609, opened 2026-06-02
Add n link pendulum regression test

moorepants:
From: https://github.com/pydy/pydy/pull/278


moorepants:
Close, this isn't needed.

PR #614, opened 2026-06-06
Add method to System for plotting state trajectories

Peter230655:
I tried to make the looks same like we have with opty.

### DIFF
--- a/pydy/system.py
+++ b/pydy/system.py
@@ -158,13 +158,17 @@ from itertools import repeat
 
 import numpy as np
 import sympy as sm
-import matplotlib.pyplot as plt
 from sympy.physics.mechanics import dynamicsymbols, find_dynamicsymbols
 from scipy.integrate import odeint
 from scipy.optimize import root
 
 from .codegen.ode_function_generators import generate_ode_function
 from .utils import PyDyFutureWarning, PyDyUserWarning
+
+plt = sm.external.import_module('matplotlib.pyplot',
+                                import_kwargs={'fromlist': ['']},
+                                catch=(RuntimeError,))
+
 
 SYMPY_VERSION = sm.__version__
 
@@ -1586,17 +1590,17 @@ class System(object):
         constants.remove(dynamicsymbols._t)
         return constants
 
-
-    def plot_trajectories(self, result, axes=None):
+    def plot_state_trajectories(self, x, axes=None):
         """Returns the axes for a plot. The plot displays the state
         trajectories versus time.
 
         Parameters
         ==========
-        result : ndarray, (len(times), num_states )
+        x : ndarray, (len(times), num_states)
             The solution obtained from pydy.system.System.integrate or a
             similar array of state trajectories.
-        axes : ndarray of AxesSubplot, shape(num_states, )
+        axes : ndarray of AxesSubplot,
+            shape(len(eom_method.q) + len(eom_method.u), )
             An array of matplotlib axes to plot to.
 
         Returns
@@ -1605,6 +1609,10 @@ class System(object):
             A matplotlib axes with the state trajectories plotted.
 
         """
+
+        if x.shape[0] != len(self.times):
+            raise ValueError('x must have the same number of rows as '
+                             'len(self.times).')
 
         num_plots = len(self.eom_method.q[:]) + len(self.eom_method.u[:])
         if axes is None:
@@ -1616,11 +1624,10 @@ class System(object):
                 raise ValueError(f'axes must have shape ({num_plots},).')
 
         for i, name in enumerate(self.eom_method.q[:] + self.eom_method.u[:]):
-            axes[i].plot(self.times, result[:, i])
+            axes[i].plot(self.times, x[:, i])
             axes[i].set_ylabel(sm.latex(name, mode='inline'))
 
-        axes[-1].set_xlabel('time')
+        axes[-1].set_xlabel('Time')
         axes[0].set_title('State Trajectories')
-
         return axes
 
--- a/pydy/system.py
+++ b/pydy/system.py
@@ -1590,7 +1590,7 @@ class System(object):
         constants.remove(dynamicsymbols._t)
         return constants
 
-    def plot_state_trajectories(self, x, axes=None):
+    def plot_states(self, x, axes=None):
         """Returns the axes for a plot. The plot displays the state
         trajectories versus time.
 
@@ -1611,8 +1611,7 @@ class System(object):
         """
 
         if x.shape[0] != len(self.times):
-            raise ValueError('x must have the same number of rows as '
-                             'len(self.times).')
+            raise ValueError('x.shape[0] = len(self.times) is violated.')
 
         num_plots = len(self.eom_method.q[:]) + len(self.eom_method.u[:])
         if axes is None:
--- a/pydy/system.py
+++ b/pydy/system.py
@@ -158,6 +158,7 @@ from itertools import repeat
 
 import numpy as np
 import sympy as sm
+import matplotlib.pyplot as plt
 from sympy.physics.mechanics import dynamicsymbols, find_dynamicsymbols
 from scipy.integrate import odeint
 from scipy.optimize import root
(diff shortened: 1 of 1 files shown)


pydy/system.py
```diff
@@ -1584,3 +1585,42 @@ def _Kane_constant_symbols(self, exprs=None):
...
+        num_plots = len(self.eom_method.q[:]) + len(self.eom_method.u[:])
+        if axes is None:
+            fig, axes = plt.subplots(num_plots, 1, sharex=True,
+                                     layout='compressed',
+                                     figsize=(6.4, 0.8*num_plots))
+        else:
+            if len(axes) != num_plots:
+                raise ValueError(f'axes must have shape ({num_plots},).')
+
+        for i, name in enumerate(self.eom_method.q[:] + self.eom_method.u[:]):
```
moorepants, line 1625:
```suggestion
        for i, name in enumerate(self.states):
```

pydy/system.py
```diff
@@ -1584,3 +1585,42 @@ def _Kane_constant_symbols(self, exprs=None):
         constants = unique_symbols
         constants.remove(dynamicsymbols._t)
         return constants
+
+
+    def plot_trajectories(self, result, axes=None):
```
moorepants:
```suggestion
    def plot_trajectories(self, x, axes=None):
```

moorepants: or something that makes more sense, "result" has no use meaning here.

pydy/system.py
```diff
@@ -1584,3 +1585,42 @@ def _Kane_constant_symbols(self, exprs=None):
...
+        if axes is None:
+            fig, axes = plt.subplots(num_plots, 1, sharex=True,
+                                     layout='compressed',
+                                     figsize=(6.4, 0.8*num_plots))
+        else:
+            if len(axes) != num_plots:
+                raise ValueError(f'axes must have shape ({num_plots},).')
+
+        for i, name in enumerate(self.eom_method.q[:] + self.eom_method.u[:]):
+            axes[i].plot(self.times, result[:, i])
```
moorepants: You need to have a check that makes sure x has the same length as self.times.

pydy/system.py
```diff
@@ -1584,3 +1585,42 @@ def _Kane_constant_symbols(self, exprs=None):
...
+            axes[i].set_ylabel(sm.latex(name, mode='inline'))
+
+        axes[-1].set_xlabel('time')
```
moorepants:
```suggestion
        axes[-1].set_xlabel('Time')
```

pydy/system.py
moorepants: single line spacing between methods on a class

pydy/system.py
```diff
@@ -158,6 +158,7 @@
 
 import numpy as np
 import sympy as sm
+import matplotlib.pyplot as plt
```
moorepants: matplotlib is currently an optional dependency of pydy, I'd like to keep it that way. You need to handle the case when matplotlib is not installed.

pydy/system.py
```diff
@@ -1584,3 +1585,42 @@ def _Kane_constant_symbols(self, exprs=None):
...
+            An array of matplotlib axes to plot to.
+
+        Returns
+        =======
+        axes : ndarray of AxesSubplot
+            A matplotlib axes with the state trajectories plotted.
+
+        """
+
+        num_plots = len(self.eom_method.q[:]) + len(self.eom_method.u[:])
```
moorepants, line 1616: `self.num_states` exists.

pydy/system.py
```diff
@@ -1584,3 +1585,42 @@ def _Kane_constant_symbols(self, exprs=None):
...
+        if axes is None:
+            fig, axes = plt.subplots(num_plots, 1, sharex=True,
+                                     layout='compressed',
+                                     figsize=(6.4, 0.8*num_plots))
+        else:
+            if len(axes) != num_plots:
```
moorepants, line 1622:
```suggestion
            if len(axes) != self.num_states:
```

pydy/system.py
```diff
@@ -1584,3 +1585,42 @@ def _Kane_constant_symbols(self, exprs=None):
...
+        Returns
+        =======
+        axes : ndarray of AxesSubplot
+            A matplotlib axes with the state trajectories plotted.
+
+        """
+
+        num_plots = len(self.eom_method.q[:]) + len(self.eom_method.u[:])
+        if axes is None:
+            fig, axes = plt.subplots(num_plots, 1, sharex=True,
```
moorepants, line 1618:
```suggestion
            fig, axes = plt.subplots(self.num_states, 1, sharex=True,
```

pydy/system.py
```diff
@@ -1584,3 +1585,42 @@ def _Kane_constant_symbols(self, exprs=None):
...
+                raise ValueError(f'axes must have shape ({num_plots},).')
+
+        for i, name in enumerate(self.eom_method.q[:] + self.eom_method.u[:]):
+            axes[i].plot(self.times, result[:, i])
+            axes[i].set_ylabel(sm.latex(name, mode='inline'))
+
+        axes[-1].set_xlabel('time')
+        axes[0].set_title('State Trajectories')
+
+        return axes
```
moorepants, line 1631: no dead spaces after this line

pydy/system.py
moorepants:
```suggestion
    def plot_state_trajectories(self, result, axes=None):
```

moorepants:
This needs unit tests and all examples that make this plot should be changed to use this method.

pydy/system.py
Peter230655: I saw it, but num_states includes the noncontributing forces if there are any,
This is why I did it like this.

pydy/system.py
Peter230655: self.states seems to include the reaction forces.

pydy/system.py
Peter230655: will  change.

pydy/system.py
Peter230655: will do it.

pydy/system.py
Peter230655: done

pydy/system.py
Peter230655: done

pydy/system.py
Peter230655: I moved it to the method.

pydy/system.py
Peter230655: done

pydy/system.py
Peter230655: done

pydy/system.py
moorepants: It does not include the reaction forces. It includes the integral of the reaction forces. These are states as currently designed. I explained this is detail in PR #583.

pydy/system.py
moorepants: That is not the correct approach.

Peter230655:
> This needs unit tests and all examples that make this plot should be changed to use this method.

Do you mean in this PR?

moorepants:
> Do you mean in this PR?

Absolutely, I do not merge any changes without tests. Same in sympy and opty.

pydy/system.py
Peter230655: > It does not include the reaction forces. It includes the integral of the reaction forces. These are states as currently designed. I explained this is detail in PR #583.

This is what I meant, sorry. In any case is is different from eom_method.q[:] + eom_method.u[:]

pydy/system.py
Peter230655: What else can I do, other than making the keyword axes mandatory?

Peter230655:
I made the changes suggested, except:
- num_states, I do not think is right as it contains the integrated noncontr. forces
- no good idea what to do with matplotlib. I could make axes mandatory, but in my view this would be contrary to the ease pydy should offer to the user.

I think, the latter must be clarified before I can change the unit tests and the examples.

pydy/system.py
Peter230655: > matplotlib is currently an optional dependency of pydy, I'd like to keep it that way. You need to handle the case when matplotlib is not installed.

Just curiosity: 
In opty we have matplotlib as a dependency since we are plotting things in opty.
Now that we want to plot similar things in pydy, why should matplotlib be avoided?

moorepants: matplotlib is an optional dependency in opty also

Peter230655: > matplotlib is an optional dependency in opty also

in opty/direct_collocation.py I saw:
```
plt = sm.external.import_module('matplotlib.pyplot',
                                import_kwargs={'fromlist': ['']},
                                catch=(RuntimeError,))

```
I do not understand this line, to be frank. Is this what we should be here, too?
chatGPT (who else can I ask?) told me this has something to do with optional dependency so I am trying it

moorepants: You don't have to ask chatgpt, you can just read the docstring for the function:

https://github.com/sympy/sympy/blob/1991693a3052bd551121f79e8129a1b22e21f4ad/sympy/external/importtools.py#L49

Peter230655:
I made all the changes suggested, except for num_states, which I do not feel is correct here.
Added the matplotlib import as copied from opty/direct_collocation.
If this is right, I will adapt the tests and the examples.

pydy/system.py
Peter230655: > You don't have to ask chatgpt, you can just read the docstring for the function:
> 
> https://github.com/sympy/sympy/blob/1991693a3052bd551121f79e8129a1b22e21f4ad/sympy/external/importtools.py#L49

Thanks!
I will be 71 years old in a few days! :-(

Peter230655:
Just changed the name of the method to ``plot_states``
If you think, the way matplotlib is handled (straight copy from opty) is correct, and the name will remain, I will start to look at the test and the examples.

Issue #610, opened 2026-06-03
Add plotting methods to System()


moorepants:
You should be able to do something like:

```python
x = system.integrate()
ax = system.plot_state_trajectories(x)
```

to get a default time history plot of the motion. We should look to opty, as it has such methods and we've working out some of the ideas there already. There are also ideas laid out in https://github.com/pydy/pydy/pull/353 which should be reviewed and taken int account.

Similar to opty, these should be basic and give a default plot that can quickly be generated.

Peter230655:
My initial thinking is:

- system.plot_state_trajectories(x)
- system.plot_noncontributing_forces(x)
- system.plot_constraint_violations(x)
- system.plot_outputs(x)

moorepants:
The plotting methods should map to the `evaluate_*` methods and take the respective inputs for those.

Peter230655:
I am unsure I understand what you mean.

moorepants:
We have several `evalaute_*` methods now. The plotting methods should match these (at least as a start).

You write `system.plot_noncontributing_forces(x)` but `x` does not contain the noncontributing forces.

Peter230655:
Clear!

moorepants:
If we make plot functions that directly map to the numerical evaluation functions,  this is what they would look like:

- `x = integrate()`: `plot_states(x=, t=)` 
- `xd = evaluate_ode(x=, t=)`: `plot_state_derivatives(xd=, t=)`
- `y = evaluate_outputs(x=, t=)`: `plot_outputs(y=, t=)`
- `fc = evaluate_constraints(x=, t=)`: `plot_constraints(fc=, t=)`
- `fh = evaluate_config_constraints(x=, t=)`: `plot_config_constraints(fc=, t=)`
- `fn = evaluate_motion_constraints(x=, t=)`: `plot_motion_constraints(fh=, t=)`

moorepants:
All of the `evaluate_*` methods can also return values for a single instance of time. Should we supporting plotting this somehow to be consistent?

Peter230655:
> If we make plot functions that directly map to the numerical evaluation functions, this is what they would look like:
> 
> * `x = integrate()`: `plot_states(x=, t=)`
> * `xd = evaluate_ode(x=, t=)`: `plot_state_derivatives(xd=, t=)`
> * `y = evaluate_outputs(x=, t=)`: `plot_outputs(y=, t=)`
[...]

PR #615, opened 2026-06-06
plot noncontributing forces

Peter230655:
I tried to have the looks like we have with opty.

### DIFF
--- a/pydy/system.py
+++ b/pydy/system.py
@@ -158,6 +158,7 @@ from itertools import repeat
 
 import numpy as np
 import sympy as sm
+import matplotlib.pyplot as plt
 from sympy.physics.mechanics import dynamicsymbols, find_dynamicsymbols
 from scipy.integrate import odeint
 from scipy.optimize import root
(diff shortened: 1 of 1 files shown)


moorepants:
I already mentioned in #610 that we need to start with plotting methods that match the `evaluate*` methods. This is not one of those. I am not sure we will even add a special one for noncontributing loads as it should shown in a `plot_output_trajectories()` method already.

Peter230655:
> I already mentioned in #610 that we need to start with plotting methods that match the `evaluate*` methods. This is not one of those. I am not sure we will even add a special one for noncontributing loads as it should shown in a `plot_output_trajectories()` method already.

I am working on a method to plot the outputs. I thought it may be convenient to the user to have a plotting routine for the noncontributing forces, as likely this will be the bulk of the output. My thinking is that for convenience we should have
- plot state trajectories
- plot noncontributing forces
- plot constraint violations

yes, the latter two would be in plot outputs.

moorepants:
I will accept methods to plot what we already "evaluate". After those are present, we can discuss if we want more special plotting functions.

Peter230655:
I see your point. I am playing around with a ``plot_outputs`` method.
Simple idea is that the user specifies what he wants plotted from the outputs available.

My problem right now is matplotlib.

PR #623, opened 2026-06-10
Set dependent initial conditions now uses nonlinear solver for coordinates and linear solver for speeds.

moorepants:


moorepants:
Just noting that I may should change the dep_vars kwarg to dep_q and dep_u (separate them). Not sure though.

